PR #4, opened 2017-07-28
Add libpng

RobLoach:
Used in [org.libretro.RetroArch](https://github.com/flathub/org.libretro.RetroArch). Pretty handy.


TingPing:
One problem is we might need 8 versions for all the popular ABI breaks they have.

TingPing:
So I'll be closing this given:

- There are only two applications on flathub using 1.2 (and one using 1.6)
- As mentioned versions are not inter-compatible so multiple versions here
- It isn't a particularly complex build

PR #5, opened 2017-08-30
Update Qt5 to latest release

AsavarTzeth:


Barthalion:
Did you have a chance to test it locally against some of existing applications, let's say mgba?

AsavarTzeth:
I just tested io.mgba.mGBA and it works fine, by which I mean it runs.

By the way, as far as I remember you would need to update your submodule anyhow for this to be an issue. For example, currently io.mgba.mGBA is locked to 21ea7b3732419017ba8e7bec1dc83a8042127fc1, meaning it would not be affected.

If I am wrong or you still worry about compatibility, may I suggest you use the LTS version, which fitting is 5.9.x at the moment. Alternatively, when the next release is due we provide a -lts version in addition to the existing one.

Barthalion:
Yeah, I was mostly curious what happens when you bump that submodule. If it builds, I'm happy to merge it (even more so that it is LTS). Thanks!

PR #7, opened 2017-09-15
Add protobuf, qt5-qtsvg, qt5-qttools

subpop:
Used by an upcoming flatpak I'm submitting, io.github.Cockatrice.cockatrice.

### DIFF
--- a/protobuf/protobuf.json
+++ b/protobuf/protobuf.json
@@ -1,0 +1,19 @@
+{
+  "name": "protobuf",
+  "cleanup": [
+    "/bin",
+    "/doc",
+    "/include",
+    "/lib/*.a",
+    "/lib/*.la",
+    "/lib/debug",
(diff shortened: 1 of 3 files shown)


TingPing:
The purpose of this repo is very commonly used libraries, you should bundle whatever you need in your own project for now unless there is wider use.

subpop:
I figured these 3 libraries are common enough to be included in a repository like this. They don't add any baggage to app repositories that don't use them, and they do allow us to define a consistent version of, say, Qt5 across all apps in the repository. Either way, I'll pull protobuf into my app PR and out of this one.

TingPing:
We base the usage on packages that already exist not ones that potentially will. AFAIK yours is the first to use these.

subpop:
Ahhhh okay. That makes more sense.

PR #11, opened 2018-03-04
Add cld2 to shared modules

cosimoc:
CC @nedrichards


nedrichards:
Great, can you put up a PR to poedit to use this as well?

PR #12, opened 2018-03-23
Add modules crawled from recipes in https://github.com/flathub

amtlib-dot-dll:
BTW, these code are "copied" from recipes in Flathub, so is there any copyright issues?

### DIFF
--- a/libXfont.json
+++ b/libXfont.json
@@ -1,0 +1,16 @@
+{
+  "name": "libXfont",
+  "cleanup": [
+    "*"
+  ],
+  "config-opts": [
+    "--disable-static"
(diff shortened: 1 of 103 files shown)


TingPing:
This isn't meant to contain every lib ever made. Just a few particularly common ones.

amtlib-dot-dll:
From https://github.com/flatpak/flatpak-builder/issues/8:
> through the hell of finding dependency sources, propper build settings and fixing compilation errors.

So where can we easily find dependency sources?

TingPing:
You build them. Once you start bundling you become a packager, that is just how it is.

PR #14, opened 2018-04-20
Add boost, gtk2 and libpng

gasinvein:
Boost or GTK2 is already used by many apps available from flathub. I think it's a good idea to provide them as reusable shared modules.
And libpng12 is needed by some proprietary apps like WPS.

### DIFF
--- a/boost/boost-configure.sh
+++ b/boost/boost-configure.sh
@@ -1,0 +1,11 @@
+
+./bootstrap.sh "$@"
+
+
+cat <<EOF >Makefile
+all:
+	./b2 -j `nproc`
+
+install:
+	./b2 install
+EOF
--- a/boost/boost.json
+++ b/boost/boost.json
@@ -1,0 +1,17 @@
+{
+    "name": "boost",
+    "buildsystem": "autotools",
+    "cleanup": [ "/lib/libboost_*.a", "/include/boost" ],
+    "sources": [
+        {
+            "type": "archive",
+            "url": "https://dl.bintray.com/boostorg/release/1.64.0/source/boost_1_64_0.tar.gz",
+            "sha256": "0445c22a5ef3bd69f5dfb48354978421a85ab395254a26b1ffb0aa1bfd63a108"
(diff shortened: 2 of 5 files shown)


TingPing:
- boost: :-1: It has too many options that may be unique to a specific project
- gtk: :-1: Maybe but the GNOME runtime already has it and the number of gtk2 applications isn't that large and should only get smaller over time
- libpng12: :-1: With a quick search we only have two applications on flathub using 1.2.

barthalion:
+1 for GTK2, not really for the rest.

ghisvail:
> With a quick search we only have two applications on flathub using 1.2.

I have got another one coming. How many apps depending on libpng12 would you need to accept?

ghisvail:
> I have got another one coming.

The application is `org.openttd.OpenTTD`. So far, the `libpng` module is declared in the `modules` metadata of the main manifest.

TingPing:
@ghisvail Why does an open source application that is still maintained depend upon libpng from 2001?

ghisvail:
Hang on that's not right. Should have been libpng16 (currently used for the
openttd built on Debian) not libpng12. My bad.

TingPing:
@ghisvail Yea 1.6 is already in the sdk and you don't have to build it.

ghisvail:
Thanks, I'll drop the superfluous libpng12 module from the manifest then.

RobLoach:
Speaking of libpng12, I had suggested that over at https://github.com/flathub/shared-modules/pull/4 .... One thing we could do is move them over to https://github.com/casept/flatpak-modules as an alternative.

Then there would be a common space for the modules that could live outside of the shared-modules space.

TingPing:
> One thing we could do is move them over to https://github.com/casept/flatpak-modules as an alternative.

I do not want another submodule repo used on Flathub though. It adds an extra layer of trust to validate.

TingPing:
GTK2 has been added and the others won't be as discussed.

PR #17, opened 2018-06-02
add lame 3.100

eszlari:


TingPing:
All files here need to be JSON.

TingPing:
Also we should probably remove the old version from here, and could you go around to the [8](https://github.com/search?l=JSON&q=org%3Aflathub+lame&type=Code) projects using lame and update it after this merges?

eszlari:
> Also we should probably remove the old version from here, and could you go around to the 8 projects using lame and update it after this merges?

That was my plan.

eszlari:
> All files here need to be JSON.

Is this going to change in the future?

TingPing:
> Is this going to change in the future?

No. JSON is a supported format that manifests will be in on Flathub and JSON is fully compatible with YAML. It makes no sense to have YAML here.

TingPing:
Now that I think about it due to how `flatpak-builder` parses the files I suppose it works fine to mix them doesn't it? In that case maybe it makes sense but for now I would prefer everything here being JSON.

eszlari:
> I suppose it works fine to mix them doesn't it?

YAML as a module works: https://github.com/flathub/io.github.quodlibet.QuodLibet

> for now I would prefer everything here being JSON

Why?

TingPing:
It is just more simple and what everything else here uses. It avoids annoying things like whitespace being off which isn't caught in CI since this repo has none.

I know you love YAML but it is as much of a pain as a pleasure in my eyes.

TingPing:
Not going to do this?

eszlari:
It turns out, git submodules are a bit messy, so I will stop using them.

PR #20, opened 2018-06-28
Add pygame 1.9.3 and dependencies

takluyver:
This follows from [my other pull request](https://github.com/flathub/flathub/pull/424) where I tried to make a pygame baseapp, but I was told that it was more appropriate as a shared module. So, here it is. :-)

I've currently left four new dependencies of pygame as internal modules, but I'm happy to split these out as separate modules if that's better. These are audiofile, libmikmod, fluidsynth and portmidi.

### DIFF
--- a/pygame/pygame-1.9.3.json
+++ b/pygame/pygame-1.9.3.json
@@ -1,0 +1,90 @@
+{
+    "name": "pygame",
+    "sources": [
+        {
+            "type": "archive",
+            "url": "https://pypi.python.org/packages/61/06/3c25051549c252cc6fde01c8aeae90b96831370884504fe428a623316def/pygame-1.9.3.tar.gz",
+            "sha256": "751021819bdc0cbe5cbd51904abb6ff9e9aee5b0e8955af02284d0e77d6c9ec2"
+        },
+        {
+            "type": "patch",
+            "path": "pygame-add-search-dirs.patch"
+        }
+    ],
+    "buildsystem": "simple",
+    "build-commands": [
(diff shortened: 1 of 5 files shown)


pygame/pygame-1.9.3.json
```diff
@@ -0,0 +1,90 @@
...
+                {
+                    "type": "patch",
+                    "path": "fluidsynth-no-rawmidi.patch"
+                }
+            ]
+        },
+        {
+            "name": "portmidi",
+            "buildsystem": "cmake-ninja",
+            "config-opts": [
```
TingPing: Add `-DCMAKE_BUILD_TYPE=Release` do all cmake projects.

pygame/pygame-1.9.3.json
```diff
@@ -0,0 +1,90 @@
...
+            "sha256": "751021819bdc0cbe5cbd51904abb6ff9e9aee5b0e8955af02284d0e77d6c9ec2"
+        },
+        {
+            "type": "patch",
+            "path": "pygame-add-search-dirs.patch"
+        }
+    ],
+    "buildsystem": "simple",
+    "build-commands": [
+      "/usr/bin/pip3 install --ignore-installed --no-deps --prefix=/app ."
```
TingPing: Don't use absolute path in the off chance somebody installs custom `pip3`.

takluyver:
Thanks, fixed up those points. :-)

barthalion:
I think it's fine to keep dependencies internal for now. Thanks!

takluyver:
Thanks!

PR #22, opened 2018-07-03
qt4 url has changed

cpba:


AsciiWolf:
There is already a fix for this in #21. ;-)

PR #24, opened 2018-07-14
Add README info on how to use shared-modules

takluyver:
This is just copied from https://github.com/flathub/flathub/wiki/App-Requirements , but I can never remember which page that info is on, so I think it would be good to have it here in the README as well.


jurf:
I’d add a link in case the wiki gets is updated before this README.

takluyver:
@jurf good plan, done

jurf:
OK, looks fine by me.

PR #28, opened 2018-08-22
Add python2 shared module

alatiera:
`org.freedesktop.Platform` has dropped `python 2` but steam just started depending on it for some reason.

I copied the cleanup stuff from [here][2] and the fd.o 18.08 python2 [cleanups][3]

[2]: https://github.com/flatpak/freedesktop-sdk-base/issues/20#issuecomment-375896426
[3]: https://gitlab.com/freedesktop-sdk/freedesktop-sdk/blob/05d273306abb0225d2d2905c7bdd8de519d87066/elements/base/python2.bst#L40-55

### DIFF
--- a/python2.7/python-2.7.15.json
+++ b/python2.7/python-2.7.15.json
@@ -1,0 +1,42 @@
+{
+    "name": "python-2.7",
+    "sources": [
+        {
+            "type": "archive",
+            "url": "https://www.python.org/ftp/python/2.7.15/Python-2.7.15.tgz",
+            "sha256": "18617d1f15a380a919d517630a9cd85ce17ea602f9bbdc58ddc672df4b0239db"
+        }
+    ],
+    "config-opts": [
+        "--enable-shared",
+        "--with-ensurepip=yes",
+        "--with-system-expat",
+        "--with-system-ffi",
+        "--enable-loadable-sqlite-extensions",
+        "--with-dbmliborder=gdbm",
+        "--enable-unicode=ucs4"
+    ],
+    "cleanup": [
+        "/bin/2to3*",
+        "/bin/easy_install*",
+        "/bin/idle*",
+        "/bin/pyvenv*",
+        "/bin/pydoc*",
+        "/include",
+        "/share",
+
+        /* Test scripts */
+        "/lib/python*/test",
+        "/lib/python*/*/test",
+        "/lib/python*/*/tests",
+        "/lib/python*/lib-tk/test",
+        "/lib/python*/lib-dynload/_*_test.*.so",
+        "/lib/python*/lib-dynload/_test*.*.so",
+
+        /* Unused modules */
+        "/lib/python*/idlelib",
+        "/lib/python*/tkinter*",
+        "/lib/python*/turtle*",
+        "/lib/python*/lib2to3*"
+    ]
+}


alatiera:
@alexlarsson though so and suggested it

TingPing:
Worth noting my cleanup list was clearly for Python3, so you have lib2to3 and stuff in the list. Please do verify it gets everything for Python 2 and remove some of the extras.

I can easily foresee this being used by a few projects so it might not be a terrible idea.

alexlarsson:
Yeah, i think with 18.08 not having python 2 we need to have a simple story for app developers who would be afraid of something like "package your own python".

alatiera:
@TingPing I think its clean now, feel free to point anything I missed.

python-2.7.15/python-2.7.15.json
```diff
@@ -0,0 +1,46 @@
...
+    "name": "python-2.7",
+    "sources": [
+        {
+            "type": "archive",
+            "url": "https://www.python.org/ftp/python/2.7.15/Python-2.7.15.tgz",
+            "sha256": "18617d1f15a380a919d517630a9cd85ce17ea602f9bbdc58ddc672df4b0239db"
+        }
+    ],
+    "buildsystem": "autotools",
+    "confgi-opts": [
```
barthalion: Should be `config-opts`.

python-2.7.15/python-2.7.15.json
```diff
@@ -0,0 +1,46 @@
+{
+    "name": "python-2.7",
+    "sources": [
+        {
+            "type": "archive",
+            "url": "https://www.python.org/ftp/python/2.7.15/Python-2.7.15.tgz",
+            "sha256": "18617d1f15a380a919d517630a9cd85ce17ea602f9bbdc58ddc672df4b0239db"
+        }
+    ],
+    "buildsystem": "autotools",
```
barthalion: It's the default.

python-2.7.15/python-2.7.15.json
```diff
@@ -0,0 +1,46 @@
...
+    "confgi-opts": [
+        "--enable-shared",
+        "--with-ensurepip=yes",
+        "--with-system-expat",
+        "--with-system-ffi",
+        "--enable-loadable-sqlite-extensions",
+        "--with-dbmliborder=gdbm",
+        "--enable-unicode=ucs4"
```
barthalion: Technically you could also pass `--enable-optimizations --enable-lto` but I've tested it only on x86_64. Does it use shared zlib too?

python-2.7.15/python-2.7.15.json
```diff
@@ -0,0 +1,46 @@
...
+        "/lib/python*/*/test",
+        "/lib/python*/*/tests",
+        "/lib/python*/lib-tk/test",
+        "/lib/python*/lib-dynload/_*_test.*.so",
+        "/lib/python*/lib-dynload/_test*.*.so",
+
+        /* Unused modules */
+        "/lib/python*/idlelib",
+        "/lib/python*/tkinter",
+        "/lib/python*/turtle*",
```
barthalion: I can see that at least Tkinter.pyo, Tkinter.pyc and Tkinter.py are still there, same about turtle.*.

python-2.7.15/python-2.7.15.json
```diff
@@ -0,0 +1,46 @@
...
+    "cleanup": [
+        "/bin/2to3*",
+        "/bin/easy_install*",
+        "/bin/idle*",
+        "/bin/python*-config",
+        "/bin/pyvenv*",
+        "/bin/pydoc*",
+        "/include",
+        "/share",
+        "/lib/pkgconfig",
```
barthalion: pkgconfig, python2.7-config and headers should probably stay as someone may want to build a module against it

python-2.7.15/python-2.7.15.json
alatiera: I used the same config we use to build py2 in fd.o Sdk, we could try to enable those but haven't tested them.

> Does it use shared zlib too?
Yes, the zlib is in the .Platform of 18.08

python-2.7.15/python-2.7.15.json
alatiera: Weird, I see none of those, I added a `tkinter*` just to make sure though.

python-2.7.15/python-2.7.15.json
alatiera: done

python-2.7.15/python-2.7.15.json
alatiera: removed

alatiera:
I also renamed the module folder to match the naming scheme of `lua5.3`

barthalion:
Thanks!

PR #29, opened 2018-08-23
Python2: Fix libpython2.7.so.1.0 permissions

alatiera:
While trying to build Steam, flatpak fails to strip
libpython2.7.so.1.0 due to libpython2.7 being installed as read-only.

Related log: https://flathub.org/builds/#/builders/2/builds/3780


nanonyme:
Looks good to me. I don't have any superpowers over this repo though. Ping @alexlarsson @TingPing

PR #35, opened 2018-10-20
Fix qt4 with Freedesktop runtime 18.08

### LINKED ISSUE #32: qt4 module does not build with org.freedesktop.Sdk 18.08
The current revision of the runtime:
org.freedesktop.Sdk/x86_64/18.08                      flathub  597122efa9e3

It should be easy to reproduce.

Unrud:
Patches from ArchLinux

fixes #32


TingPing:
Have you tested that it still works against 1.6?

Unrud:
``qt4-openssl-1-1.patch`` breaks compatibility with older versions.

TingPing:
modules here will have to work on both for a while.

TingPing:
It looks like the patch *tries* to check for openssl version and work on both.

Unrud:
It causes type conversion errors because of the added ``const``.

PR #37, opened 2018-11-20
Add dbus-glib module

wjt:
Based on the block in @ramcq's Tomboy manifest at https://github.com/flathub/flathub/pull/722, reformatted to YAML for legibility and updated to dbus-glib 0.110.

dbus-glib appears in 5 existing apps on Flathub (not including Tomboy).

### DIFF
--- a/dbus-glib/dbus-glib-0.110.json
+++ b/dbus-glib/dbus-glib-0.110.json
@@ -1,0 +1,23 @@
+{
+    "name": "dbus-glib",
+    "cleanup": [
+        "*.la",
+        "/bin",
+        "/etc",
+        "/include",
+        "/libexec",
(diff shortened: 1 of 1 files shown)


TingPing:
Everything here is in JSON.

wjt:
Does it have to be?

wjt:
But okay, I have a `json-to-yaml` script from the days before flatpak-builder supported YAML so I'll run it.

TingPing:
I'll leave porting apps to use this up to you also =)

wjt:
Okay, that's all of them with the exception of https://github.com/flathub/org.libreoffice.LibreOffice which I think is a copy of the manifest in LibreOffice Git.

PR #42, opened 2019-01-07
glu: --disable-static

bam80:


jurf:
I guess `*.a` can be removed from the `cleanup` now?

PR #54, opened 2019-03-18
Add gtkmm module

wjt:
These are the latest stable versions at the time of writing. The rules
are derived from org.gnome.Gnote, which is one of 10 apps already on
Flathub which uses these libraries.

(Perhaps we should consider making an SDK extension for this?)

### DIFF
--- a/gtkmm/gtkmm.json
+++ b/gtkmm/gtkmm.json
@@ -1,0 +1,125 @@
+{
+    "name": "gtkmm",
+    "sources": [
+        {
+            "type": "archive",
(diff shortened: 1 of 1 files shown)


TingPing:
> (Perhaps we should consider making an SDK extension for this?)

I kind of like that idea personally. These still get regular updates and honestly don't look that hard to build. The main benefit people would want is avoiding build time and bumping versions right?

wjt:
> I kind of like that idea personally. These still get regular updates and honestly don't look that hard to build. The main benefit people would want is avoiding build time and bumping versions right?

Indeed. I don't have time to work on this any further at the moment, so if you'd prefer not to add this to shared-modules I'll just go and inline it into Gnote and pavucontrol.

TingPing:
Probably best for now, they would just stagnate here.

PR #57, opened 2019-04-02
add systemd-udev

Pixelnarium:
Since udev-175 is broken right now and probably will also break in the future I think it is a good idea to offer an alternative to the udev sources from 2012.
I saw that [the flatpaked Steam](https://github.com/flathub/com.valvesoftware.Steam) pulls udev from systemd which is maintained. I tried it with my packages and it seems to work quite well (within the boundaries of flatpak of course).

I don't know the rules of shared-modules but I still ask myself if the filename itself should be versioned. udev-175 is... other shared-modules are not (i.e. glew). systemd updates come quite often but I think the udev code is not changed that much but that is just a guess. I think it should not be versioned to reduce the maintenance burden of packages who are using it. But in this PR it is versioned to be in line with udev-175. The version also comes with the shared-modules submodule commit reference automatically. New versions are not shoved into every packages automatically and untested.

Maybe also asking @nanonyme and @LeandroStanger (who are working on this in the Steam Flatpak) what they think about this. Maybe I missed something here.


LeandroStanger:
@ramcq @gasinvein This will be useful for Lutris Flatpak

LeandroStanger:
@RobLoach @gasinvein This will be useful for Lutris Flatpak

udev/systemd-udev-241.json
```diff
@@ -0,0 +1,79 @@
+{
+  "name": "hwdata",
+  "config-opts": [
+    "--datarootdir=/app/share"
+  ],
+  "sources": [
+    {
+      "url": "https://github.com/vcrhonek/hwdata/archive/v0.321.tar.gz",
```
LeandroStanger: Update to hwdata v0.321

udev/systemd-udev-241.json
```diff
@@ -0,0 +1,79 @@
+{
+  "name": "hwdata",
```
gasinvein: hwdata not needed for systemd.

Pixelnarium: So it is not part of the systemd-hwdb stuff? It looked like something udev uses. Sure, will remove if it is totally unrelated and unneeded.

Erick555:
Maybe [this](https://github.com/gentoo/eudev) would be better alternative for bringing udev which doesn't need to compile whole systemd?

eszlari:
> Maybe this would be better alternative for bringing udev which doesn't need to compile whole systemd?

But most parts can be disabled in meson. And I'm sure upstream (being mostly Redhat people) would be open to add more options (if needed) to support the flatpak use case.

Pixelnarium:
> Maybe [this](https://github.com/gentoo/eudev) would be better alternative for bringing udev which doesn't need to compile whole systemd?

As far as I can see the end result is pretty much just the udev library even if it compiles a little bit more, it is deleted afterwards. Same with eudev since it also a little more than just libudev.so. I think in the end it is just a matter of taste and I have no personal preference on that. I just took the approach from the Steam flatpak to start the discussion. If more people want to use eudev it is fine by me.

Erick555:
Flatpak usecase is to make udev as dumb as possible as most of it's functionality won't work anyway therefore I think eudev is better suited for. the job.

> As far as I know flatpak also includes some parts of systemd. There is a libsystemd.so in the runtime.

It was added to be able to run runtime as standalone system in vm. For flatpak it's useless.

eszlari:
> I think eudev is better suited for the job.

Depends on have well that fork is maintained. I think it makes more sense to work on something like
```
meson -Dudev_only_build=true
```
 for systemd than rely on some fork.

udev/systemd-udev-241.json
gasinvein: usbutils and pciutils uses it. Systemd has it's own [copy](/systemd/systemd/tree/master/hwdb) of `pci.ids` and `usb.ids.

Pixelnarium: Okay. I removed it.

Pixelnarium:
It was already in there the whole time. Just thought that hwdata is part of this.

gasinvein:
@Pixelnarium  Oh, sorry, didn't notice it.

Erick555:
> Depends on have well that fork is maintained. I think it makes more sense to work on something like

it's critical part of most non-systemd distros so it have to be maintained well 😄 

> meson -Dudev_only_build=true

Then why it's not used here?

eszlari:
> Then why it's not used here?

I meant that such a option could be added to systemd's meson build system.

gasinvein:
Is this still WIP?

Pixelnarium:
The filename ist pretty much the last question (except from updating systemd itself).

Pixelnarium:
That's my offer for now.

TingPing:
This seems fine, if a bit much, but I'd like to see this tested by some current udev module users in practice to ensure everything is good.

gasinvein:
@TingPing it's currently used by Steam, see flathub/com.valvesoftware.Steam#178.

Erick555:
@gasinvein I think that an app used old udev should be tested against the new one.

Lctrs:
I tested [GitKraken](https://github.com/flathub/com.axosoft.GitKraken) with a custom local build of the [Electron2 base app](https://github.com/flathub/org.electronjs.Electron2.BaseApp) including systemd-udev this morning and it worked well.

Erick555:
@Lctrs speaking of GitKraken, you may be interested [in this fixes for git build](https://github.com/flathub/org.qownnotes.QOwnNotes/pull/9).

jjardon:
FYI, freedesktop-sdk 19.08 will include libudev (from systemd)

Erick555:
@TingPing freedesktop-sdk 19.08 was released with libudev. This can be closed.

Issue #56, opened 2019-04-01
udev broke again


Pixelnarium:
Fails to build again
```
/usr/lib/gcc/x86_64-unknown-linux-gnu/8.3.0/../../../../x86_64-unknown-linux-gnu/bin/ld: udev/udevd.o: in function `static_dev_create_from_modules':
/run/build/udev/udev/udevd.c:898: undefined reference to `makedev'
/usr/lib/gcc/x86_64-unknown-linux-gnu/8.3.0/../../../../x86_64-unknown-linux-gnu/bin/ld: udev/udevd.o: in function `is_devpath_busy':
/run/build/udev/udev/udevd.c:517: undefined reference to `major'
/usr/lib/gcc/x86_64-unknown-linux-gnu/8.3.0/../../../../x86_64-unknown-linux-gnu/bin/ld: /run/build/udev/udev/udevd.c:540: undefined reference to `major'
/usr/lib/gcc/x86_64-unknown-linux-gnu/8.3.0/../../../../x86_64-unknown-linux-gnu/bin/ld: udev/udev-event.o: in function `udev_event_apply_format':
/run/build/udev/udev/udev-event.c:233: undefined reference to `major'
/usr/lib/gcc/x86_64-unknown-linux-gnu/8.3.0/../../../../x86_64-unknown-linux-gnu/bin/ld: /run/build/udev/udev/udev-event.c:241: undefined reference to `minor'
/usr/lib/gcc/x86_64-unknown-linux-gnu/8.3.0/../../../../x86_64-unknown-linux-gnu/bin/ld: /run/build/udev/udev/udev-event.c:357: undefined reference to `major'
/usr/lib/gcc/x86_64-unknown-linux-gnu/8.3.0/../../../../x86_64-unknown-linux-gnu/bin/ld: /run/build/udev/udev/udev-event.c:374: undefined reference to `minor'
/usr/lib/gcc/x86_64-unknown-linux-gnu/8.3.0/../../../../x86_64-unknown-linux-gnu/bin/ld: /run/build/udev/udev/udev-event.c:374: undefined reference to `major'
/usr/lib/gcc/x86_64-unknown-linux-gnu/8.3.0/../../../../x86_64-unknown-linux-gnu/bin/ld: /run/build/udev/udev/udev-event.c:385: undefined reference to `minor'
[...]

PR #58, opened 2019-04-02
fix udev build

Pixelnarium:
Multiple files now need `#include <sys/sysmacros.h>`. Seems to work. Tried it with one game that needs udev.

Fixes https://github.com/flathub/shared-modules/issues/56


Erick555:
ping @TingPing @barthalion

gasinvein:
Just of curiosity, why did this happen?

Pixelnarium:
Updates inside the runtime. Maybe GCC. It was updated to 8.3 [lately](https://gitlab.com/freedesktop-sdk/freedesktop-sdk/merge_requests/1041).

Erick555:
@gasinvein this keeps happening because this udev version come from 2012 and it's less and less likely that it will still build with modern tools. That's why finding replacement is needed.

PR #59, opened 2019-04-03
Bump to 2.7.16

Erick555:
This is also the version used in Sdk.


jurf:
Well, you have to manually update the submodule to get the version, so breaking the manifest shouldn’t be a problem.

Erick555:
Currently you have to update submodules **AND** change manifest which is annoying for maintainers. I don't see the reason for breaking manifest on every version bump unless you provide one.

PR #73, opened 2019-09-28
add glu 9.0.1

user5145:

PR #74, opened 2019-09-28
update glu to version 9.0.1

user5145:
I'm sending #73 as another pr to keep glu's history.


Erick555:
IMO it should be renamed to `glu/glu-9.json` to prevent breaking app manifests on every minor bump.

PR #77, opened 2019-10-19
Python cleanups and IBus update

bochecha:
This:

* adds a few files to the cleanup list for the python-2.7 shared module;
* updates IBus to its latest release in the gtk2 share module;

### DIFF
--- a/gtk2/gtk2.json
+++ b/gtk2/gtk2.json
@@ -100,8 +100,8 @@
       "sources": [
         {
           "type": "archive",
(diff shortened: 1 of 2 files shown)


Erick555:
@bochecha @barthalion @TingPing 
Considering this will [break ibus for fdo 18.08 again](https://github.com/flathub/shared-modules/commit/c0bdf798e31da54a8f8bcb25ca17ad957515e1ca#diff-6f8c1b9f39a89f8b77368e53ed4c8e10) maybe we should really split ibus to separate submodule and version it per runtime like `ibus-gtk2-18-08.json`, `ibus-gtk2-19-08.json` and so on...

PR #78, opened 2019-10-20
Replace udev with gudev

Erick555:
Since freedesktop-sdk 19.08 (gnome 3.34 and kde 5.13) udev is part
of runtime. Some gnome/gtk apps may need additionaly gudev which
was part of old udev shared-module and it's not part of runtime.

This change will intentinally break manifests of apps that used
udev from shared-modules. The assumption is that app using
freedesktop 18.08 runtime won't need updating shared-modules and
if they do then it's always recommended to switch to 19.08 at the
same time where apps can either  stop bundling udev or replace it
with gudev. Currently apps often move to the new runtime without
being aware that the don't need udev anymore.

gudev manifest was based on:
https://github.com/flathub/org.mixxx.Mixxx/blob/0b2e80bcce2f4e4d89a0e166829c7b5da8385bb4/org.mixxx.Mixxx.yaml#L286

It's also similar to how gnome runtime builds it (but not expose
to flatpak yet):
https://gitlab.gnome.org/GNOME/gnome-build-meta/blob/3.34.1/elements/core-deps/libgudev.bst


Erick555:
ping @TingPing @barthalion

PR #86, opened 2020-01-16
Add libsecret

Lctrs:
Same reasons/motivations as for openjpeg

See https://github.com/search?q=org%3Aflathub+libsecret&type=Code

### DIFF
--- a/libsecret/libsecret.json
+++ b/libsecret/libsecret.json
@@ -1,0 +1,23 @@
+{
+  "name": "libsecret",
+  "buildsystem": "meson",
+  "config-opts": [
+    "-Dmanpage=false",
+    "-Dvapi=false",
+    "-Dgtk_doc=false"
+  ],
+  "cleanup": [
+    "/bin",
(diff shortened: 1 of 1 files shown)


TingPing:
Seems reasonable, do you plan on updating the places that use it?

libsecret/libsecret.json
```diff
@@ -0,0 +1,24 @@
...
+    "-Dvapi=false",
+    "-Dgtk_doc=false"
+  ],
+  "cleanup": [
+    "/bin",
+    "/include",
+    "/lib/pkgconfig",
+    "/share/gir-1.0",
+    "/share/man",
+    "*.la"
```
TingPing:
```suggestion
```

Lctrs:
> Seems reasonable, do you plan on updating the places that use it?

Yes.

libsecret/libsecret.json
Lctrs: Done.

Issue #85, opened 2019-12-01
Set up code owners for modules


TingPing:
For details on what code owners is:

- https://github.blog/2017-07-06-introducing-code-owners/
- https://help.github.com/en/github/creating-cloning-and-archiving-repositories/about-code-owners

As shared-modules grows and gets more security sensitive modules we should have maintainers for each module so there is somebody taking responsibility for what is here rather than it becoming a dumping ground.

(Brought up because of #84)

Proposal for maintainers (need volunteers):

- [ ] SDL
- [ ] cld2
- [x] dbus-glib (@TingPing)
- [ ] glew
- [ ] glu
- [x] gtk2 (@TingPing)
- [x] gudev (@Erick555)
- [x] intltool (@TingPing)
- [ ] lame
- [x] libappindicator (@TingPing) 
- [ ] libmad
- [x] libsecret (@Lctrs)
- [ ] lua
- [x] python (@bilelmoussaoui)
- [ ] pygame
- [ ] qt4
- [ ] smpeg
- [x] openjpeg (@bochecha)

Then in the future all new modules will require an owner to be added.

Erick555:
I can take care of gudev and hope [it could go away at some point in future](https://gitlab.gnome.org/GNOME/gnome-build-meta/issues/196).

Lctrs:
As I proposed to introduce it, I can take care of `libsecret`.

TingPing:
Merged though volunteers would still be appreciated for some modules.

bilelmoussaoui:
I can take care of python2

Lctrs:
The `CODEOWNERS` file is currently useless for a big majority of modules if we don't have write access to this repository.

See https://github.com/flathub/shared-modules/blob/master/CODEOWNERS with all the errors highlighted.

bochecha:
Why is openjpeg an error ? It seems to be assigned to me, what's the problem ?

-- 
Bochecha

Le 6 juin 2022 12:29:43 GMT+02:00, Lctrs ***@***.***> a écrit :

Lctrs:
@bochecha that's because you don't have write access to this repository, as neither of the users highlighted in red does. So when a PR is made for these modules, we don't get a notification for it and can't merge it.

bochecha:
Ah right !

It's because I was eventually forced to change my github account from bochecha to mbridon, because having lost the 2FA I lost access to my account and the github support team was unable to help :-/

So I renamed it from my nickname to my real name.

[...]

PR #91, opened 2020-02-11
Set up code owners for some modules

bochecha:
As shared-modules grows and gets more security sensitive modules we
should have maintainers for each module so there is somebody taking
responsibility for what is here rather than it becoming a dumping
ground.

While we don't have yet owners for each module, some people already came
forward to commit to maintain some of them.

This change adds those people as owners of those modules, to begin
making some progress on #85.

---

I thought about making all unowned modules owned by the Flathub team, but not being a member of the Flathub organization I apparently can't see how the teams are called. :man_shrugging:

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,0 +1,10 @@
+# Fallback on Flathub admins for unowned shared modules
+*                  @flathub/reviewers
+
+/dbus-glib/        @TingPing
+/gtk2/             @TingPing
+/gudev/            @Erick555
+/intltool/         @TingPing
+/libappindicator/  @TingPing
+/libsecret/        @Lctrs
+/openjpeg/         @bochecha


CODEOWNERS
```diff
@@ -0,0 +1,11 @@
+# Fallback on Flathub admins for unowned shared modules
+# TODO: What would be the right team?
+*                   @flathub/some-team
```
TingPing:
```suggestion
*                   @flathub/reviewers
```

CODEOWNERS
```diff
@@ -0,0 +1,11 @@
...
+
+dbus-glib/**        @TingPing
```
Lctrs:
```suggestion
/dbus-glib/         @TingPing
```

Same for the others below. See https://help.github.com/en/github/creating-cloning-and-archiving-repositories/about-code-owners#codeowners-syntax.

CODEOWNERS
bochecha: Done in the latest push to this branch.

CODEOWNERS
bochecha: The `**` glob would have effectively been equivalent.

Good catch on the missing leading `/` though, that could have caused real issues for nested folders with the same names. :+1:

This is now fixed in the latest push to this branch.

PR #93, opened 2020-02-14
misc: auto bump apps using shared-modules

### LINKED ISSUE #92: Auto bump the shared-modules for sensitive ones
Shared-modules contains some of the sensitive modules. So far, updating the module here with the latest security patches isn't enough as the apps have to update their shared-modules submodule and re-build the app with it.

We can setup a Github action that creates a merge request automatically for each app that uses it on Flathub. This will make it easier for app maintainers too :)

bilelmoussaoui:
This adds a Github action to automatically bump all the apps using shared-modules when this repository gets updated. It doesn't check if the used modules by the app were updated or not yet, but that shouldn't be an issue for a first run I guess.

Shouldn't be merged until we add a secret token for flathubbot 

Depends on https://github.com/flathub/actions/pull/7
Fixes #92

### DIFF
--- a/.github/workflows/auto-mr.yml
+++ b/.github/workflows/auto-mr.yml
@@ -1,0 +1,14 @@
+name: auto PR
+
+on:
+  push:
+    branches:
+      - master
+
+jobs:
+  build-deploy:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: flathub/actions/shared-modules-auto-pr@master
+        env:
+          GITHUB_TOKEN: ${{ secrets.FLATHUBBOT_TOKEN }}


Lctrs:
Also closes https://github.com/flathub/flatpak-external-data-checker/issues/76

barthalion:
Can you move it to flathub/flathub? I think it better fits "organization-wide" CI.

bilelmoussaoui:
I thought about doing so, but I wanted it to be run only when something changes on shared-modules and I don't think we can do that from an other repo

barthalion:
We should run it periodically too. In the worst case it does nothing useful.

bilelmoussaoui:
Replaced by https://github.com/flathub/flathub/pull/1372

PR #95, opened 2020-03-28
glu: provide a build for 9.0.1

aleixpol:
Fixes a build issue with freedesktop 19.08 among possibly other things.


Erick555:
@aleixpol [exact same thing](https://github.com/flathub/shared-modules/blob/master/glu/glu-9.json) already exist, what I'm missing? If you have issues during submodule update then you should update your manifest to include correct submodule filename not duplicating same modules under different filenames.

aleixpol:
I didn't see it, I don't really understand.

Sorry.

PR #96, opened 2020-04-02
libappindicator: Apply another crash fix patch

kujeger:
From discussion in https://github.com/flathub/com.discordapp.Discord/issues/30

Based on 71810457bfa88d914e7570df04d86220fbafe3ed but should apply cleanly

PR #97, opened 2020-04-02
libappindicator: Correct path in patch

kujeger:
Fix for the improper #96 .

Currently running locally built Discord on a locally built Electron2.BaseApp with this patch applied. Havn't had a crash yet.


kujeger:
@TingPing could you link me to where it's failing? I'm unable to reproduce here

TingPing:
I triggered a new build, maybe it works now: https://flathub.org/builds/#/builders/32/builds/18010

PR #99, opened 2020-04-06
gtk2: unhardcode libdir from ibus sub-module

gasinvein:
Use original makefile to install client library, and use pkg-config to determine full path to `immodules.cache`.
This would help builds where libdir is not `/app/lib`, e.g. in multilib case (those commonly set libdir to `/app/lib32`).

### DIFF
--- a/gtk2/gtk2.json
+++ b/gtk2/gtk2.json
@@ -94,7 +94,7 @@
       ],
       "post-install": [
         "gtk-query-immodules-2.0 > immodules.cache",
-        "install immodules.cache -t $(pkg-config --variable=libdir gtk+-2.0)/gtk-2.0/$(pkg-config --variable=gtk_binary_version gtk+-2.0)/"
+        "install immodules.cache -t $(pkg-config --variable=libdir gtk+-2.0)/gtk-2.0/2.10.0/"
       ],
       "sources": [
         {
--- a/gtk2/gtk2.json
+++ b/gtk2/gtk2.json
@@ -68,7 +68,6 @@
     },
     {
       "name": "ibus-gtk2",
-      "no-make-install": true,
       "config-opts": [
         "--disable-xim",
         "--disable-dconf",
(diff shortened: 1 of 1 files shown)


gtk2/gtk2.json
```diff
@@ -90,12 +90,12 @@
...
-      "ensure-writable": [
-        "/lib/gtk-2.0/2.10.0/immodules.cache"
+      "build-commands": [
+        "make -C client/gtk2 install"
       ],
       "post-install": [
-        "install -m644 --target-directory=${FLATPAK_DEST}/lib/gtk-2.0/2.10.0/immodules client/gtk2/.libs/im-ibus.so",
-        "gtk-query-immodules-2.0 > ${FLATPAK_DEST}/lib/gtk-2.0/2.10.0/immodules.cache"
+        "gtk-query-immodules-2.0 > immodules.cache",
+        "install immodules.cache -t $(pkg-config --variable=libdir gtk+-2.0)/gtk-2.0/$(pkg-config --variable=gtk_binary_version gtk+-2.0)/"
```
TingPing, line 98: I realize its more accurate, but gtk2 also hasn't had an ABI break since like 2008, I don't think we need to query the binary version.

gtk2/gtk2.json
```diff
@@ -90,12 +90,12 @@
         "--disable-introspection",
         "--disable-python2"
       ],
-      "ensure-writable": [
-        "/lib/gtk-2.0/2.10.0/immodules.cache"
+      "build-commands": [
+        "make -C client/gtk2 install"
```
TingPing, line 94: `make-install-args` is probably the better place for this.

gtk2/gtk2.json
gasinvein: It's a directory path. I.e. it normally becomes `/app/lib/gtk-2.0/2.10.0/`.

TingPing: Yes, my point was `2.10.0` is a string that will never change to the end of time, bit of a complication for no gain.

gasinvein: Oh, I see.

PR #102, opened 2020-04-25
Added Linux Audio section

hfiguiere:
This adds to shared modules some modules comonly found for Linux Audio. See for the base work to provide plugin support: flathub/flathub#1460 (it will need an update to point to the right repository)

- `fftw3f.json` : this is fftw3 with float (it's a separate build with a different configure option) and is found in a lot of flathub packages. I'm willing to progressively submit PR for these packages. This is not audio specific though, so I'm willing to move it in a different directory if you prefer.

- `ladspa.json`, `lrdf.json` : LADSPA plugin support. `lrdf` depends on `ladspa`, but `ladspa` is standalone otherwise.

- `lv2.json` : LV2 plugin support

- `dssi.json` , `liblo.json`: liblo provide OSC support and dssi provide DSSI support. Sometime DSSI plugin want OSC as well.

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -9,3 +9,4 @@
 /libsecret/        @Lctrs
 /openjpeg/         @bochecha
 /python2.7/        @bilelmoussaoui
+/linux-audio/      @hfiguiere
--- a/linux-audio/dssi.json
+++ b/linux-audio/dssi.json
@@ -1,0 +1,21 @@
+{
+    "name": "dssi",
+    "rm-configure": true,
+    "config-opts": [
+        "--disable-static"
+    ],
+    "sources": [
+        {
+            "type": "archive",
+            "url": "http://download.sf.net/sourceforge/dssi/dssi-1.1.1.tar.gz",
+            "sha256": "f2c82b073a947c8255284249097667f9b14e660bf86186f3fcd3b3b3e087814e"
+        },
+        {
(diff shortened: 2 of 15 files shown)


Erick555:
> fftw3f.json : this is fftw3 with float (it's a separate build with a different configure option) and is found in a lot of flathub packages. I'm willing to progressively submit PR for these packages. This is not audio specific though, so I'm willing to move it in a different directory if you prefer.

Maybe you can ask freedesktop-sdk to enable float in runtime version?

hfiguiere:
> Maybe you can ask freedesktop-sdk to enable float in runtime version?

They actually try to remove stuff instead of adding. I'm just making it easier to build stuff.

Erick555:
I don't understand, freedesktop-sdk does ship fftw right now so if it's unusable by some apps then devs would be interested in fixing it. Not having to build something in fist place is better than making it easier to build.

hfiguiere:
Even then, it wouldn't work right now so it still helps to have things in here.

hfiguiere:
I took ownership of that section

eszlari:
> Maybe you can ask freedesktop-sdk to enable float in runtime version?

https://gitlab.com/freedesktop-sdk/freedesktop-sdk/-/issues/444

hfiguiere:
So we still need it for the current SDK version.

PR #108, opened 2020-05-28
qt4 -- fix building with gcc9 and higher

cpba:


cpba:
I've tested this locally and there aren't any obvious issues.

I have backported this fix from qt5, to me it looks like a simple macro. However, I'm very unfamiliar with the qt codebase and can't say for sure that this won't affect anything else.

PR #110, opened 2020-06-14
Don't clean up secret-tool

mzabaluev:
It can be useful for troubleshooting (e.g. https://github.com/flathub/com.visualstudio.code/issues/159) inside the sandbox.


TingPing:
Its not generally useful, developers can modify the manifest locally to debug issues.

mzabaluev:
The tool takes 0.0034% of the installed size of `org.freedesktop.Platform`, but oh well, let's be frugal.

PR #112, opened 2020-06-15
lua5.3: Compile with hardening flags

Unrud:
Use `$CFLAGS` as defined by the SDK.
Also use symlinks for the libraries instead of copies because the loader might print a warning.

### DIFF
--- a/lua5.3/lua-5.3.5.json
+++ b/lua5.3/lua-5.3.5.json
@@ -3,7 +3,9 @@
     "buildsystem": "simple",
     "build-commands": [
         "make CFLAGS=\"$CFLAGS -fPIC -DLUA_COMPAT_5_2 -DLUA_COMPAT_5_1 -DLUA_USE_DLOPEN\" linux",
-        "make TO_LIB='liblua.so liblua.so.5.3 liblua.so.5.3.5' INSTALL_TOP=${FLATPAK_DEST} install",
+        "make TO_LIB=liblua.so.5.3.5 INSTALL_TOP=$FLATPAK_DEST install",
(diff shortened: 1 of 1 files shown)


Erick555:
I think -fPIC may be dropped as PIC/PIE is enabled in sdk gcc by default.

Unrud:
> I think `-fPIC` may be dropped as PIC/PIE is enabled in sdk gcc by default.

The build fails without it.

lua5.3/lua-5.3.5.json
```diff
@@ -1,15 +1,11 @@
...
-        "CFLAGS=-O2 -g -fPIC -DLUA_COMPAT_5_2 -DLUA_COMPAT_5_1 -DLUA_USE_DLOPEN",
-        "linux"
-    ],
-    "make-install-args": [
-        "INSTALL_TOP=${FLATPAK_DEST}",
-        "TO_LIB=liblua.so liblua.so.5.3 liblua.so.5.3.5"
-    ],
-    "post-install": [
+    "buildsystem": "simple",
+    "build-commands": [
```
TingPing, line 4: Why change the build system type?

Unrud: Because I could not use the variable `$CFLAGS` in `make-args`. Using the autotools build system also seems kind of hacky as LUA is not using autotools.

TingPing: Ok well add `-j $FLATPAK_BUILDER_N_JOBS` to the make invocations then.

Unrud: Done.

PR #114, opened 2020-06-24
Add libusb v1.0.23

A6GibKm:


libusb/libusb.json
```diff
@@ -0,0 +1,19 @@
+{
+    "name": "libusb",
+    "config-opts": [ "--disable-static", "--disable-udev" ],
```
gasinvein, line 3: Without udev, it won't have access to hwdb, and without it `lsusb` will be unable to print device names, which apps commonly rely on.

A6GibKm: Fixed it in https://github.com/flathub/shared-modules/pull/115. Thanks for the clarification.

PR #121, opened 2020-08-26
Add Monkey Audio Codec for easier management

Eonfge:
@enzo1982 Paging you on this one. 

By adding the MAC script to a shared module, multiple projects can benefit from it. Now it won't be necessary anymore to copy-paste the fre:ac script to Asunder, Flacon and others.

### DIFF
--- a/mac/mac.json
+++ b/mac/mac.json
@@ -5,6 +5,9 @@
   "make-install-args": [
     "prefix=/app"
   ],
+  "cleanup": [
+    "/include/MAC"
+  ],
   "sources": [
     {
         "type": "archive",
--- a/mac/mac.json
+++ b/mac/mac.json
@@ -1,0 +1,22 @@
+{
+  "name": "libmac",
+  "buildsystem": "simple",
+  "build-commands": [
+    "make -f Source/Projects/NonWindows/Makefile prefix=/app all install"
+  ],
+  "cleanup": [
+    "/bin/mac"
+  ],
+  "sources": [
+    {
+        "type": "archive",
+        "url": "https://freac.org/patches/MAC_SDK_544.zip",
+        "sha256": "d5d2c2de4f64c61bc98c59a8691e47a6e2bb1beb3a3ae50d98bb0fdfdb91e480",
+        "strip-components": 0
+    },
+    {
+        "type": "patch",
+        "path": "mac-sdk-5.44-gcc.patch"
+    }
+  ]
+}
--- a/mac/mac.json
+++ b/mac/mac.json
@@ -1,8 +1,9 @@
 {
   "name": "libmac",
-  "buildsystem": "simple",
-  "build-commands": [
-    "make -f Source/Projects/NonWindows/Makefile prefix=/app all install"
+  "no-autogen": true,
+  "rm-configure": true,
+  "make-install-args": [
+    "prefix=/app"
   ],
   "cleanup": [
     "/bin/mac"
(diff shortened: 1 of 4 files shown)


Eonfge:
@TingPing You mention that people shouldn't ask to include libraries that have only limited use, but I think this is a bit of an edge case: Monkey Audio Codec is a surprisingly fiddly thing to manage as a dependency, and it has a very fast release cycle. I hope that by adding it to  shared module, it becomes easier to manage for the multiple projects that require it.

If you approve it, and @enzo1982 is willing to share maintenance on this module, then that would be great

TingPing:
Add your usernames to `CODEOWNERS` as well.

mac/mac.json
```diff
@@ -0,0 +1,22 @@
+{
+  "name": "libmac",
+  "buildsystem": "simple",
+  "build-commands": [
+    "make -f Source/Projects/NonWindows/Makefile prefix=/app all install"
```
TingPing, line 5: You can probably just use `make-install-args` and `no-autogen: true`.

Eonfge: I'm just the messenger in this case: I just copy-paste the Fre:ac implementation every time, and this shared-module is supposed to streamline that a bit.

Eonfge:
I've added me and @enzo1982 to the code owners, assuming that Enzo wants to share responsibility on this.

enzo1982:
Alright, let's do this! Thanks @Eonfge for getting this started!

I prepared an updated patch for the current Monkey's Audio version 5.54. @Eonfge please give me access to this fork and I will amend this PR. I'll also change the build instructions as suggested by @TingPing then.

I'm ready to take care of updating this module regularly if it is accepted and will make the fre:ac build script use it too.

Eonfge:
Invite send. Cool to have you on board :sunglasses:

enzo1982:
Thanks! I pushed the update to 5.54.

mac/mac.json
```diff
@@ -0,0 +1,23 @@
+{
+  "name": "libmac",
+  "no-autogen": true,
+  "rm-configure": true,
+  "make-install-args": [
+    "prefix=/app"
+  ],
+  "cleanup": [
+    "/bin/mac"
+  ],
```
Eonfge, line 10: Not sure if this cleanup should be included though. Since we try to have a generic package here, some applications might need the main binary to use mac.

**Edit.** Can confirm. Removing "/bin/mac" actually breaks mac support in Asunder. @enzo1982 are you OK with removing the cleanup?

enzo1982: You're right, of course. I removed that command.

Eonfge:
@TingPing Well, all good to go!

mac/mac.json
```diff
@@ -0,0 +1,20 @@
+{
+  "name": "libmac",
+  "no-autogen": true,
+  "rm-configure": true,
+  "make-install-args": [
+    "prefix=/app"
+  ],
```
TingPing, line 7: Surely this is missing all the `"cleanup"` for left over headers and whatnot.

enzo1982: Alright, 67f6320 adds the `cleanup` for headers. The `install` target installs only `/bin/mac`, `/lib/libMAC.so` and `/include/MAC/`.

Eonfge: Good for me. Works in the test I have

TingPing: Is the binary `mac` used often?

Eonfge: - ca.littlesvr.asunder
- com.github.Flacon

Both need it. Not sure if that counts as much, but otherwise it would kind of invalidate all this work. The whole point behind making this shared-modules, is so I and enzo can share some work on some very finicky codecs.

TingPing: Just checking. Thanks!

PR #122, opened 2020-09-13
Add libhandy

A6GibKm:
Also added older libhandy v0.13.0 since many packages depend on it and it should
require zero maintenance.

- Note that this probably won't work for Glade since it sets `glade_catalog=false`, I don't know if there is a real reason to remove the catalog for `libhandy-1` in any case.

- I am debating if 0.13.0 should use the old Purism source or the newer gitlab.gnome.org source. @bilelmoussaoui Do you think GNOME could add a source for libhandy-0 under [https://download.gnome.org/sources/libhandy/](download.gnome.org/sources/libhandy)?

While I can maintain this, I don't have the time right now to send PRs to the 51+ repos who use some version of libhandy acording to [this search](https://github.com/search?q=org%3Aflathub+libhandy&type=code). I guess it only makes sense to push to repos using versions V satisfying 0.13.0 < V =< 1.0.0, which should not be that many. When I have the time I will remove the draft status.

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -6,6 +6,7 @@
 /gudev/            @Erick555
 /intltool/         @TingPing
 /libappindicator/  @TingPing
+/libhandy/         @A6GibKm
 /libsecret/        @Lctrs
 /libusb/           @A6GibKm
 /openjpeg/         @bochecha
--- a/libhandy/libhandy-0.json
+++ b/libhandy/libhandy-0.json
@@ -1,0 +1,18 @@
+{
+    "name" : "libhandy-0",
+    "buildsystem" : "meson",
+    "config-opts" : [
+        "-Dexamples=false",
+        "-Dglade_catalog=disabled",
+        "-Dintrospection=disabled",
+        "-Dtests=false",
+        "-Dvapi=false"
+    ],
+    "sources" : [
+        {
+            "type": "archive",
+            "url": "https://gitlab.gnome.org/GNOME/libhandy/-/archive/v0.0.13/libhandy-v0.0.13.tar.gz",
+            "sha256": "645355a009f23f254eaec7752b9489c3c2f5832397fcec75433a7e00efbfe52f"
+        }
+    ]
+}
--- a/libhandy/libhandy-1.json
+++ b/libhandy/libhandy-1.json
@@ -1,0 +1,18 @@
+{
+    "name" : "libhandy-1",
+    "buildsystem" : "meson",
+    "config-opts" : [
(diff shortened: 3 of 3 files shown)


bilelmoussaoui:
I don't think it makes sense to have libhandy as a shared module. It's pretty easy to build and will probably end up in the GNOME SDK in the future as more and more of GNOME apps uses it.

bilelmoussaoui:
cc @alatiera @Exalm

Exalm:
Also by passing `-Dintrospection=disabled` and `-Dvapi=false` you're ensuring only C (and maybe Rust?) apps can use it, but not Python, JavaScript or Vala.

A6GibKm:
@bilelmoussaoui yes, I agree that it is easy. But on the other hand if you look on the list of apps that use it, most of them are in versions <0.8. The use of shared-modules helps maintainer to not have to know the status of each dependency.

> will probably end up in the GNOME SDK in the future as more and more of GNOME apps uses it.

If this is the case for 3.38-3.40, it might not worth it to add this to shared-modules.

A6GibKm:
> Also by passing -Dintrospection=disabled and -Dvapi=false you're ensuring only C (and maybe Rust?) apps can use it, but not Python, JavaScript or Vala.

Noted, will remove on next commit.

bilelmoussaoui:
> @bilelmoussaoui yes, I agree that it is easy. But on the other hand if you look on the list of apps that use it, most of them are in versions <0.8. The use of shared-modules helps maintainer to not have to know the status of each dependency.
> 
> > will probably end up in the GNOME SDK in the future as more and more of GNOME apps uses it.
> 
> If this is the case for 3.38-3.40, it might not be worth it to add this to shared-modules.

Flatpak is about letting maintainers manage the specific dependencies to their application. Most of the libraries in shared-modules are there because they are either difficult to build, requires a bunch of patches or could cause security issues.

libhandy just doesn't fit what shared modules were made for.

A6GibKm:
Before I close this PR, I would like to ask what are the plans for the inclusion of libhandy on the runtime/Sdk 3.38, if any.

Exalm:
3.38 definitely no, it's deep in freeze, nightly/3.40 maybe.

A6GibKm:
Ok, I was half-expecting that, thanks.

alatiera:
No I t most luckily won't make it to the GNOME Runtime. But regardless I don't see much of a point to having a shared module. If apps are using pre-1.0 version they need to update to 0.13 anyway and nothing will change there, ever. And for the rest 1.0 seems fine.

PR #125, opened 2020-10-11
libsecret: update to 0.20.4

Lctrs:


A6GibKm:
I just discovered that `-Dgcrypt=false` solves a long-standing issue with https://github.com/flathub/ch.protonmail.protonmail-bridge. What is the implication of this flag?

Lctrs:
From https://gitlab.gnome.org/GNOME/libsecret/-/blob/master/meson_options.txt, `With gcrypt and transport encryption`. So I would rather not disable it.

A6GibKm:
Yeah, I was just wondering if maybe there is a secure fallback. For the moment I will keep using 19.1 instead of upgrading to 20.4 without `gcrypt`.

Lctrs:
@A6GibKm `com.microsoft.Teams` is hit by this bug. I've been thinking about the fix you proposed and I'm wondering about its validity. My guess is that maybe the secret got encoded twice, once by the lib in the flatpak before passing it to the host and once again by the host. WDYT ?

A6GibKm:
@Lctrs  I sincerely don't know, here is the relevant issue

https://gitlab.gnome.org/GNOME/libsecret/-/issues/49

But I suspect it might be on the portal side. For the moment I am just sticking with 19.1 for protonmail-bridge, removing gcrypt might be a bad idea, but I don't understand libsecret enough to know if it is as bad as it sounds.

A6GibKm:
Does anyone remember why it was ok at the time to remove introspection? It looks wrong... is libsecret in the GNOME runtime?

A6GibKm:
Ok, then its fine. Thanks.

PR #129, opened 2020-12-23
Update libusb to 1.0.24

A6GibKm:
I will test tomorrow.


A6GibKm:
Tested with cc.arduino.arduinoide and it worked.

PR #131, opened 2021-01-01
New openjpeg

tsdgeos:


Erick555:
@tsdgeos @TingPing nothe that openjpeg2 is part of freedesktop runtime since 20.08 so app should move away from using shared-module. You may even consider removing it.

tsdgeos:
But by using the freedekstop runtime one i would be using an obsolete 2.3.1 instead of 2.4.0 so that doesn't seem like a good thing?

hfiguiere:
how obsolete is it?

tsdgeos:
100 commits, various CVEs

Erick555:
@tsdgeos I don't think shared-modules are meant for providing just a new version of software that is already in runtime. The only reason openjpeg is in runtime is to stop apps having to bundle it so if apps don't want use it from runtime then it has no purpose.

Runtime did [backport of cve fixes](https://gitlab.com/freedesktop-sdk/freedesktop-sdk/-/merge_requests/2539) and you may ask to backport more if something was missed.

tsdgeos:
> The only reason openjpeg is in runtime is to stop apps having to bundle it so if apps don't want use it from runtime then it has no purpose.

So you're saying that users should just use old versions of things because someone decided this library belongs into the runtime?

That makes 0 sense, i want my users to be use the last version of things, not some 2 year old version. If I wanted that i would just recommend them to run debian instead of flathub packages.

Erick555:
@tsdgeos 

> So you're saying that users should just use old versions of things because someone decided this library belongs into the runtime?

I'm saying that the mere fact there is new release of some software doesn't justify bundling it regardless if there are 100 or 1000 new commits unless some of them fixes a bug your app faced or add a feature you were waiting for. I'm not myself a fan of openjpeg in runtime but if it's there then ignoring it without specific reason makes everything worse. If you are convinced that having non-latest openjpeg is not acceptable for apps then you may discuss this with freedesktop maintainers.

> That makes 0 sense, i want my users to be use the last version of things, not some 2 year old version. If I wanted that i would just recommend them to run debian instead of flathub packages.

This argument applies to every single library in the runtime and effectively kills the the idea shared runtime for flatpak apps (in current form). Following this logic apps would bundle everything which bloats their sizes and adds maintenance burden.

I think flathub goal is to deliver latest **apps** (which is what most users care about) alongside debian-like stable platform. I don't really see why do you think old app + old library is better than new app + old library.

tsdgeos:
I'm going to stop answering here, this is a silly discussion, if you want to remove openjpeg from here fine, i'll add it to okular and that's it.

Erick555:
The fact you shipped vulnerable openjpeg for months while patched version was available in runtime shows best what was silly here.

tsdgeos:
Please, in the future try not to call me or my colleagues silly. It's not nice being attacked because we just disagree on things.

Erick555:
I didn't call you or your colleagues silly. I just re-used the word you used first. to show how it feels. You can't throw words towards others if you don't want to hear them back.

tsdgeos:
Let me try explain the difference that i see.

I called *the discussion* silly, because in my opinion, it had reached a non productive state, we will do what we think it is best, it's clear you disagree, to keep arguing about if we should paint the shed blue or red is silly, just wastes both our time. As you can see I didn't call you silly.

On the other hand you said our decision to use what openjpeg upstream ships was silly, i hope you understand I feel this is a somewhat more personal attack than saying the discussion is silly.

tsdgeos:
Anyhow https://gitlab.com/freedesktop-sdk/freedesktop-sdk/-/issues/1186 should be doable

Erick555:
You called discussion in which I played main part _silly_ which was demeaning for me while I participated in it with good faith I tried to make it pure technical. I didn't think it was non-productive at that point otherwise I wouldn't continue it and I don't agree with your interpretation.

> On the other hand you said our decision to use what openjpeg upstream ships was silly, i hope you understand I feel this is a somewhat more personal attack than saying the discussion is silly.

You describe your feelings while don't taking into account mine. You say I wasn't insulted because you say so and you was insulted because you say so which I may mirror here but it would lead nowhere for real.

I think what can be accepted by both of us is either: insults were mutual or there were none. Everything else would be unfair for one side.

tsdgeos:
> I think what can be accepted by both of us is either: insults were mutual or there were none. Everything else would be unfair for one side.

Ok.

TingPing:
I have removed openjpeg from here. Indeed the goal of this repo is not to provide newer versions than runtimes, the goal was to share maintenance burden for things not in the runtime.

While it is unfortunate that the runtime in this case was outdated that is still the better place for this to be maintained and they *should* be responsive for issues like CVEs (they at least have paid maintainers).

To state the obvious you are free to bundle newer versions in applications if that maintenance burden works for you.

Erick555:
To clarify some things:

- runtime wasn't _outdated_, it had same version as shared-module which was the latest until just few days ago when upstream made new release
- runtime had cve patches from day one (shared-module didn't until recent version bump)
- AFAIK there is only one paid maintainer

PR #136, opened 2021-02-11
Add libcanberra

hadess:
It's used by about 20 different Flathub modules, and the last released
version doesn't include a particular bad bug fix which would cause
applications to assert() when used on Wayland, as they try to access
the X11 display.

### DIFF
--- a/libcanberra/libcanberra.json
+++ b/libcanberra/libcanberra.json
@@ -1,0 +1,35 @@
+{
+    "name": "libcanberra",
+    "cleanup": [
+        "*.la",
+        "/bin",
+        "/etc",
+        "/include",
+        "/libexec",
+        "/share/gtk-doc",
+        "/share/man"
+    ],
+    "config-opts": [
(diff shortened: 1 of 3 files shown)


libcanberra/libcanberra.json
```diff
@@ -0,0 +1,35 @@
...
+        "/share/man"
+    ],
+    "config-opts": [
+        "--disable-static",
+        "--disable-gtk-doc",
+        "--disable-oss",
+        "--enable-pulse",
+        "--disable-udev",
+        "--disable-gtk",
+        "--disable-gtk3",
```
hadess, line 19: Typo...

hadess:
The list of modules using libcanberra:
https://github.com/search?q=org%3Aflathub+%22libcanberra%22&type=code

hadess:
Anyone?

hfiguiere:
I'm for it. @TingPing ?

Issue #137, opened 2021-02-19
Python 2 and Gtk+2: PyGtk


Eonfge:
In another pull request, I've included a file that combines the shared-modules/gtk2 module with the shared-modules/python2 module:

https://github.com/flathub/flathub/pull/2121

The file pk2gtk2.json builds pygtk which is needed for many old UI applications who rely on both Python 2 and Gtk+2. Since both of these technologies are deprecated and multiple distributions have been busy removing their libraries, I think it would be a good addition to Flathub to have a shared-module to support these older applications in a limited, containerized fashion.

PR #138, opened 2021-02-20
Include python 2 - gtk+ 2 bridge for many old applications

Eonfge:
Mentionded in #137 
---
The file pk2gtk2.json builds pygtk which is needed for many old UI applications who rely on both Python 2 and Gtk+2. Since both of these technologies are deprecated and multiple distributions have been busy removing their libraries, I think it would be a good addition to Flathub to have a shared-module to support these older applications in a limited, containerized fashion.


TingPing:
Add yourself to `CODEOWNERS` as the maintainer of this module.

Also curious how many applications on Flathub use pygtk?

TingPing:
Can you just name it `pygtk`, thats the official upstream name.

Eonfge:
I've copied the original code from here:
https://github.com/flathub/com.github.labyrinth_team.labyrinth

And a quick scan on the pull-requests list reveals another application using pygtk:
https://github.com/flathub/flathub/pull/1779

As for currently available applications, I don't have the full overview, but I think that there are still a few out there.

PR #140, opened 2021-05-23
linux-audio: fluidsynth2 2.2.1

Eonfge:
Both dynamic and static

### DIFF
--- a/linux-audio/fluidsynth2-static.json
+++ b/linux-audio/fluidsynth2-static.json
@@ -22,8 +22,8 @@
   "sources": [
     {
       "type": "archive",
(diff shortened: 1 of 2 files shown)


Eonfge:
For your reference, I tested this with GZDoom. Not sure if that is enough testing for a shared module to be updated though, are there any hard rules on when shared-modules can get updated?

TingPing:
> are there any hard rules on when shared-modules can get updated?

Not really. If its an API break then common sense applies.

One benefit to submodules is its very easy to hold back/revert an update in a project if an issue arises. At least temporarily.

hfiguiere:
I'll test with the few audio apps later tonight. they are pretty good with API.

hfiguiere (approved):

hfiguiere:
Thanks!

PR #142, opened 2021-05-31
Add GZDoom as a shared-module

Eonfge:
I've rolled into maintaining both GZDoom and the FreeDoom packages, but this is mostly just copy-pasting the same basic configuration. As such, I wish to propose a special shared module that can be used by other projects, like Blade of Agony, to share dependencies.

@Alexander-Wilms I hope you find this an interesting proposition. I've also made an upstream request to better support Flatpak in the future, although that should be of no issue in case of third-party IWADs

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -15,3 +15,4 @@
 /lua5.1/           @Unrud
 /mac/              @enzo1982 @Eonfge
 /pygtk/            @Eonfge
+/gzdoom/           @Eonfge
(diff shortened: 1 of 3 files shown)


Alexander-Wilms:
Sure, sounds good. Having flatpak-external-data-checker tags would make maintenance easier.

Eonfge:
> Having flatpak-external-data-checker tags would make maintenance easier.

I don't know if those actually work with shared modules. I've subscribed myself to all the related git projects so I'll be informed when updates are published though.

Eonfge:
@Alexander-Wilms I did have an extra look at the GZDoom bundle that you use in BoA and I don't think that you need the widescreen fix any more, these are bundled with GZDoom and when you build from git, all of that is included. An extra benefit, building from git, is that you'll also have the proper version number information in all the stack traces: the `g4.6.0` tag is fed directly into the binary.

@hfiguiere Are there any approval processes that I have to clear?

hfiguiere:
@TingPing usually deal with this.

PR #144, opened 2021-06-15
mac: update to 6.29

enzo1982:
Monkey's Audio can now be built for Linux without any patches.

I switched the build system to _simple_, because the default build system requires the Makefile in the base directory (this was previously provided by the patch). It would be great if it was possible to specify an alternative Makefile location in Flatpak manifests.


TingPing:
> because the default build system requires the Makefile in the base directory (this was previously provided by the patch).

Just set the `subdir` property.

enzo1982:
> Just set the `subdir` property.

_Just_ setting the subdir property doesn't work as the Makefile expects to be run in the base directory. When setting subdir, it becomes necessary to pass the base directory and makefile location in make-args, like this:

```
{
  "name": "libmac",
  "no-autogen": true,
  "subdir": "Source/Projects/NonWindows",
  "make-args": [
    "--directory=../../..",
    "--makefile=Source/Projects/NonWindows/Makefile"
  ],
  "make-install-args": [
    "--directory=../../..",
    "--makefile=Source/Projects/NonWindows/Makefile",
    "prefix=/app"
  ],
  "cleanup": [
    "/include/MAC"
  ],
  "sources": [
    {
        "type": "archive",
        "url": "https://freac.org/patches/MAC_SDK_629.zip",
        "sha256": "af62c05d83c003c3af8772c01ca8796b8f2ac75a2f5ce1af183ab021cd7f3cf0",
        "strip-components": 0
    }
  ]
}
```

I find this a lot uglier than just using the _simple_ build system, but if you prefer this, let me know and I'll update the PR.

mac/mac.json
```diff
@@ -1,23 +1,18 @@
 {
   "name": "libmac",
-  "no-autogen": true,
-  "rm-configure": true,
-  "make-install-args": [
-    "prefix=/app"
+  "buildsystem": "simple",
+  "build-commands": [
+    "make -f Source/Projects/NonWindows/Makefile prefix=/app all install"
```
TingPing, line 5: This is fine in that case. Just add `-j $FLATPAK_BUILDER_N_JOBS` to it.

enzo1982: Done. I had to split the `make all` and `make install` into two commands, though, as the `install` target does not have any prerequisites in the makefile so with `-j` it would be executed before everything has been built.

I'll contact the upstream author and propose some makefile changes to fix this and the other issues, so we can simplify this again in a future update.

Issue #119, opened 2020-07-11
Add Qt5 and PyQt5


darkdragon-001:
This is used in a vast number of packages. Would be a lot easier to maintain it only once.

Lctrs:
- `Qt5`: just use `org.kde.Platform`
- `PyQt5`: not really realistic IMHO. Most apps will want to enable only the modules they need to reduce build time/install size, but if you really want to, a good starting point could be what we did in [`org.musicbrainz.Picard`](https://github.com/flathub/org.musicbrainz.Picard/blob/master/org.musicbrainz.Picard.json#L36-L156). We successfully build `PyQt5` using `SIP5`.

PR #145, opened 2021-06-27
[WIP] Add PyQt and PyQtWebEngine

tinywrkb:
This adds version 5 of PyQt and PyQtWebEngine Python binding.  
I've been maintaining these modules in the qutebrowser packaging in the short period since I adopted that package, and it's I think it's better suited here.  
Calibre is an example of an app that can benefit from the PyQtWebEngine shared module, as correctly it uses the binary release and not building from source, so it's limited to x86_64.

Modules depend on the KDE 5.15 runtime.  
PyQtWebEngine also requires the [QtWebEngine base app](https://github.com/flathub/io.qt.qtwebengine.BaseApp/).

The following modules were also added as separated shared modules:

* freeglut
* python-opengl
* dbus-python

The following modules are packaged but were not separated into shared modules as they are build dependencies, are not needed on run time, used by the SIP buildsystem, so I kept them in the `sip.json` module. Removing these modules in the cleanup step will break applications that need them, so by default they must be packaged.

* python-packaging
* python-pyparsing

`x-checker-data` properties were added and `f-e-d-c` can update the modules when running on an app manifest.

For documentation purpose, I kept the commit that uses sip 4, if someone can't build with sip6 then sip4 is there in the commit history.

This was tested with qutebrowser and confirmed to build for x86_64 and aarch64 targets and run on x86_64 in flathub/org.qutebrowser.qutebrowser#54.

I haven't tried to enable every PyQt binding, I think I will leave it for a future PR when I'll have other apps to test with.

### DIFF
--- a/pyqt/pyqt5.json
+++ b/pyqt/pyqt5.json
@@ -1,8 +1,7 @@
 {
     "name": "pyqt5",
     "config-opts": [
-        "--assume-shared",
-        "--concatenate",
+        "--concatenate=1",
         "--confirm-license",
         "--enable=QtCore",
         "--enable=QtDBus",
@@ -17,28 +16,25 @@
         "--enable=QtWebChannel",
         "--enable=QtWidgets",
         "--no-designer-plugin",
-        "--no-dist-info",
         "--no-docstrings",
+        "--no-make",
         "--no-qml-plugin",
-        "--no-qsci-api",
-        "--no-stubs",
         "--no-tools",
-        "QMAKE_CFLAGS_RELEASE='-I/usr/include/python3.8/'",
-        "QMAKE_CXXFLAGS_RELEASE='-I/usr/include/python3.8/'"
+        "--api-dir=/app/share/qt/qsci/api/python",
+        "--qmake-setting=QMAKE_CFLAGS_RELEASE='-I/usr/include/python3.8/'",
+        "--qmake-setting=QMAKE_CXXFLAGS_RELEASE='-I/usr/include/python3.8/'",
+        "--qt-shared"
     ],
-    "build-options": {
-        "arch": {
-            "aarch64": {
-                "config-opts": [
-                    "--disable-feature=PyQt_Desktop_OpenGL"
-                ]
-            }
-        }
-    },
+    "make-args": [
+        "--directory=build"
+    ],
+    "make-install-args": [
+        "--directory=build"
+    ],
     "sources": [
         {
             "type": "archive",
-            "url": "https://pypi.python.org/packages/source/P/PyQt5/PyQt5-5.15.4.tar.gz",
+            "url": "https://files.pythonhosted.org/packages/8e/a4/d5e4bf99dd50134c88b95e926d7b81aad2473b47fde5e3e4eac2c69a8942/PyQt5-5.15.4.tar.gz",
             "sha256": "2a69597e0dd11caabe75fae133feca66387819fc9bc050f547e5551bce97e5be",
             "x-checker-data": {
                 "type": "pypi",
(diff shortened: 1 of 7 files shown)


Erick555:
I don't think calibre is good example of potential user as there is no plan to build it from source anytime soon.

Lctrs:
I'm -1 for PyQt for reasons explained here : https://github.com/flathub/shared-modules/issues/119

But if stated in a README that this is here to serve as a base example, then why not.

Lctrs:
Also, I would not add dbus-python as its use is discouraged. See https://dbus.freedesktop.org/doc/dbus-python/index.html#problems-and-alternatives

tinywrkb:
@Lctrs 

> PyQt5: not really realistic IMHO. Most apps will want to enable only the modules they need to reduce build time/install size,

Maybe I can get the module to read the enabled bindings list from a file supplied by the app packaging, 

_shared-modules/pyqt/pyqt5.json_

```
sources:
...
  - type: file
    path: ../../pyqt5-features.txt
```  

> Also, I would not add dbus-python as its use is discouraged. See https://dbus.freedesktop.org/doc/dbus-python/index.html#problems-and-alternatives

Alright, I'll remove it.

> I don't think calibre is good example of potential user as there is no plan to build it from source anytime soon.

I planned to try it myself, though I'm not sure I'll submit as that repo seems to be in full auto mode, and there's no integration of `flatpak-pip-generator` in `f-e-d-c`.

hfiguiere:
> I'm -1 for PyQt for reasons explained here : #119
> 
> But if stated in a README that this is here to serve as a base example, then why not.

It's funny you mention Picard, because it PyQt5 that caused the build to not be on `aarch64`.... (I have submitted a PR for that BTW).

And the trick to build PyQt5 on `aarch64` is there.

hfiguiere:
Also sip 6 is incompatible with some package that use PyQt5. 

Example: https://github.com/flathub/com.sourcepole.kadas/pull/3

(for Picard: https://github.com/flathub/org.musicbrainz.Picard/pull/77 )

pyqt/pyqt5.json
```diff
@@ -0,0 +1,115 @@
...
+        "--enable=QtSql",
+        "--enable=QtTest",
+        "--enable=QtWebChannel",
+        "--enable=QtWidgets",
+        "--no-designer-plugin",
+        "--no-docstrings",
+        "--no-make",
+        "--no-qml-plugin",
+        "--no-tools",
+        "--api-dir=/app/share/qt/qsci/api/python",
```
Lctrs, line 23: Is this line really necessary ?

pyqt/pyqt5.json
```diff
@@ -0,0 +1,115 @@
+{
+    "name": "pyqt5",
+    "config-opts": [
```
Lctrs, line 3: Please add `--debug` (generate builds with debuginfo) and `--verbose` (enable verbose/debug output during the build)

pyqt/pyqt5.json
```diff
@@ -0,0 +1,115 @@
...
+        "--enable=QtWebChannel",
+        "--enable=QtWidgets",
+        "--no-designer-plugin",
+        "--no-docstrings",
+        "--no-make",
+        "--no-qml-plugin",
+        "--no-tools",
+        "--api-dir=/app/share/qt/qsci/api/python",
+        "--qmake-setting=QMAKE_CFLAGS_RELEASE='-I/usr/include/python3.8/'",
+        "--qmake-setting=QMAKE_CXXFLAGS_RELEASE='-I/usr/include/python3.8/'",
```
Lctrs, line 25: By default, it won't be built using the default compilation flags defined in the runtime. You'll need to rely on an external file as done in https://github.com/flathub/org.musicbrainz.Picard as flatpak doesn't expand environment viariables in `config-opts` AFAIK.

pyqt/pyqt5.json
tinywrkb: I actually can't remember. I've done the sip6 conversion a couple of months back, and left it in WIP state until got back to it today.  
I'll retest this.

Lctrs: According to the [doc](https://www.riverbankcomputing.com/static/Docs/sip/command_line_tools.html#cmdoption-sip-build-api-dir), it's useless in our case. You can just drop it.

TingPing:
Add yourself to `CODEOWNERS`

pyqt/pyqt5.json
tinywrkb: What about apps that depend on QScintilla? Without this `PyQt5.api` will not be installed.

Lctrs: 1. You're cleaning it on https://github.com/flathub/shared-modules/pull/145/files#diff-e0d667a45d00fde2ccbdd1f55b4b238ea7836804c284e4b89c20fc03a9c7f8afR67 anyway.
2. I don't think most apps will need it. We're back to why I think this is not a good idea to have PyQt as a shared-modules.

tinywrkb: > You're cleaning it on...

The logic was that it's only needed when packaging an app, not for running.

> We're back to why I think this is not a good idea to have PyQt as a shared-modules.

The fact that you're here pointing out issues and helping improve the packaging of this module, illustrates the importance of having this as a shared module that is more discoverable than a module that hides in some random app packaging.

pyqt/pyqt5.json
```diff
@@ -0,0 +1,115 @@
...
+                        "name": "PyQt-builder"
+                    }
+                }
+            ],
+            "cleanup": [
+                "*"
+            ]
+        },
+        {
+            "name": "pyqt5-sip",
```
Lctrs, line 96: Can also be entirely cleaned.

pyqt/sip.json
```diff
@@ -0,0 +1,78 @@
...
+            "type": "archive",
+            "url": "https://files.pythonhosted.org/packages/f8/b2/fcd5e964eefce0737512fb4ea263308769c671c3b1b9b1e380a5008ffef0/sip-6.1.1.tar.gz",
+            "sha256": "52d25af2fcd764c4e15cc9cd1350bdb0e63f52dfa2aa3c5e7679af7fde9f7e20",
+            "x-checker-data": {
+                "type": "pypi",
+                "name": "sip"
+            }
+        }
+    ],
+    "cleanup": [
```
Lctrs, line 19: Can also be entirely cleaned.

pyqt/sip.json
```diff
@@ -0,0 +1,78 @@
...
+        "/bin"
+    ],
+    "modules": [
+        {
+            "name": "python-pyparsing",
```
Lctrs, line 24: Can also be entirely cleaned.

pyqt/sip.json
```diff
@@ -0,0 +1,78 @@
...
+                    "sha256": "c203ec8783bf771a155b207279b9bccb8dea02d8f0c9e5f8ead507bc3246ecc1",
+                    "x-checker-data": {
+                        "type": "pypi",
+                        "name": "pyparsing"
+                    }
+                }
+            ]
+        },
+        {
+            "name": "python-packaging",
```
Lctrs, line 42: Can also be entirely cleaned.

pyqt/sip.json
```diff
@@ -0,0 +1,78 @@
...
+                    "sha256": "5b327ac1320dc863dca72f4514ecc086f31186744b84a230374cc1fd776feae5",
+                    "x-checker-data": {
+                        "type": "pypi",
+                        "name": "packaging"
+                    }
+                }
+            ]
+        },
+        {
+            "name": "python-toml",
```
Lctrs, line 60: Can also be entirely cleaned.

tinywrkb: The app might need the module. Even if the packaging have another module which installs python-toml, and the cleanup phase of the module here doesn't remove the files of the added one, pip's default behavior is to not install a Python module if it's already exist.  
This means that the `--ignore-installed` option should be passed to pip, but the problem is that with `flatapk-pip-generator` the maintainer don't have control on pip's options.  
For example, KDE screwed this up with the Pygments module, where it installed in the SDK but not in the Platform runtime.

TingPing:
All of the dependencies you are also building I don't think should be toplevel shared modules. They should just be private in the pyqt directory.

kevinsmia1939:
Hi,
Any update to this?
Because there are some package that I want to use this pyqt5 submodule for because it is a bit of pain to maintain.

tinywrkb:
I'm not pushing forward with the shared module approach. 
Instead, I'm creating a PyQtWebEngine base app. I started working on this [here](https://github.com/tinywrkb/flathub/tree/com.riverbankcomputing.PyQtWebEngine.BaseApp), but was a bit sidetracked.   
I'll have something ready this week, hopefully.  
We can have cleanup script for removing PyQtWebengine and QtWebengine for those who just PyQt.

tinywrkb:
Closed in favor of flathub/flathub#2998.

PR #149, opened 2021-08-08
Update to GLU 9.0.2

rderooy:
Also use git tag instead of archive

### DIFF
--- a/glu/glu-9.json
+++ b/glu/glu-9.json
@@ -4,8 +4,8 @@
   "sources": [
     {
       "type": "archive",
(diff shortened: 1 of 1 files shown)


TingPing:
> Also use git tag instead of archive

Why? It just slows down every build and makes it use more space?

rderooy:
I figured it would be preferred for a git based repository. And it is more descriptive, it link directly to the commit tag.

Am I to assume that git tags are not preferred then?

TingPing:
Its mostly my personal preference to avoid them. Using tarballs just speeds up building, and since shared-modules are shared slowing things down is only multiplied.

Anyway, its a minor detail.

PR #150, opened 2021-08-10
Update fluidsynth to 2.2.2

rderooy:


Eonfge:
@TingPing Is this one you can pick up? Guess that @hfiguiere Is occupied.

hfiguiere:
sorry I got distracted and it was skipped :-(

Eonfge:
No problem. There are like 10 active maintainers here, responsible for running what's essentially an entire Linux Distribution. We'll miss some stuff on occasional :shrug:

PR #164, opened 2021-11-16
Make with only one job on libappindicator.

HugLifeTiZ:
Hi, all. I made a PR to an application that uses libappindicator, and the bot build failed. @ebassi on Matrix determined the cause of the build failure to be a parallelization issue within libappindicator. As the library is unlikely to see any real changes to properly fix that, this PR works around it by building the module without parallelization with `make -j1`.


HugLifeTiZ:
Thanks for the quick merge! ❤

PR #166, opened 2021-11-26
Add SDL_gfx module

wjt:
This requires special care and feeding to build on aarch64, which I
borrowed from org.manaplus.ManaPlus when org.tuxpaint.Tuxpaint gained a
dependency on this library.

I suspect (without proof) that this library is one of the reasons that
the other existing user of SDL_gfx on Flathub,
com.roguetemple.HyperRogue, is not built for aarch64.

### DIFF
--- a/SDL/SDL_gfx-2.0.25.json
+++ b/SDL/SDL_gfx-2.0.25.json
@@ -1,0 +1,24 @@
+{
+    "name": "SDL_gfx",
+    "config-opts": [
+        "--disable-static"
+    ],
+    "build-options": {
+        "arch": {
+            "aarch64": {
+                "//": "Code assumes that if __GNUC__ is defined and MMX is enabled, then x86-specific mmintrin.h is available",
+                "config-opts": [
+                    "--disable-static",
+                    "--disable-mmx"
+                ]
+            }
+        }
+    },
+    "sources": [
+        {
+            "type": "archive",
+            "url": "https://downloads.sourceforge.net/project/sdlgfx/SDL_gfx-2.0.25.tar.gz",
+            "sha256": "556eedc06b6cf29eb495b6d27f2dcc51bf909ad82389ba2fa7bdc4dec89059c0"
+        }
+    ]
+}


SDL/SDL_gfx-2.0.25.json
```diff
@@ -0,0 +1,29 @@
...
+        "--disable-static",
+        "--disable-mmx"
+    ],
+    "build-options": {
+        "arch": {
+            "i386": {
+                "config-opts": [
+                    "--enable-mmx"
+                ]
+            },
```
hfiguiere, line 14:
```suggestion
```

arch is no longer supported.

hfiguiere: I would actually remove the `disable-mmx` from config opts and just add it for `aarch64`. I feels less convoluted.

and I would submit a patch upstream to actuall properly detect architectures.

hfiguiere:
For HyperRogue: https://github.com/flathub/com.roguetemple.HyperRogue/pull/1

SDL/SDL_gfx-2.0.25.json
wjt: I understand your point. FWIW I did it this way around because the `--enable-mmx` code will fail to compile on any non-x86 platform, so when someone comes along and adds RISC-V to Flathub, another arch-specific block would need to be added.

hfiguiere: when it happens, we'll see. Maybe by that time they'll have apply the patch I recommend submitting.

wjt: I have to say I am not filled with enthusiasm at the prospect of sending a patch upstream. SDL_gfx (this one) appears to be abandoned; SDL2_gfx, its SDL2 successor, only marginally less so.

SDL/SDL_gfx-2.0.25.json
```diff
@@ -0,0 +1,24 @@
...
+    "name": "SDL_gfx",
+    "config-opts": [
+        "--disable-static"
+    ],
+    "build-options": {
+        "arch": {
+            "aarch64": {
+                "//": "Code assumes that if __GNUC__ is defined and MMX is enabled, then x86-specific mmintrin.h is available",
+                "config-opts": [
+                    "--disable-static",
```
hfiguiere, line 11: AFAIK the config-opts are cumulative.

TingPing:
I'd like somebody added to CODEOWNERS if they are willing.

LaxentaInc:
@wjt https://colorwall.xyz how is this project i am working on peeps @hfiguiere @TingPing @orowith2os @ebassi

LaxentaInc:
idk what better way to ask than to tell devs idk

PR #170, opened 2021-12-24
Added lua5.4

hfiguiere:
Darkatable needs it.

See https://github.com/flathub/org.darktable.Darktable/pull/81


TingPing:
Please have somebody in `CODEOWNERS`

hfiguiere:
tbf I know nothing about Lua. All I know is that Darktable was using 5.3 before and now require 5.4, so I thought it might be worth it.

TingPing:
Small enough, I'll deal with it.

hfiguiere:
thanks!

PR #171, opened 2021-12-28
pygtk: cleanup all devel and demo artifacts

yselkowitz:
/cc @Eonfge


Eonfge:
Ow, thanks for this! I did a quick compare with one of the projects I maintain and this saves about 16 MB!

PR #173, opened 2022-02-01
Add Vorbisgain

Eonfge:
From a reliability point-of-view, we should really ensure that all packages using Vorbisgain adopt this shared module, which has all the patches from Debian. Upsteam doesn't seem to be working on it any longer.

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -16,3 +16,4 @@
 /mac/              @enzo1982 @Eonfge
 /pygtk/            @Eonfge
 /gzdoom/           @Eonfge
+/vobisgain/        @Eonfge
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -16,4 +16,4 @@
 /mac/              @enzo1982 @Eonfge
 /pygtk/            @Eonfge
 /gzdoom/           @Eonfge
-/vobisgain/        @Eonfge
+/vorbisgain/       @Eonfge
--- a/vorbisgain/0001-temp_files.patch
+++ b/vorbisgain/0001-temp_files.patch
@@ -1,0 +1,75 @@
+Description: Apply patch to use temp files, which are dependent
+ on the file which is beeing processed, instead always using the same
+ filename, which can result to data loss in scenarios where two or more
+ vorbis processes are running parallel.
+Author: Pavel N. Krivitsky <pavel@krivitsky.name>
+Bug-Debian: http://bugs.debian.org/cgi-bin/bugreport.cgi?bug=505164
+Forwarded: no
+---
+ vorbis.c |   17 +++++++++--------
+ 1 file changed, 9 insertions(+), 8 deletions(-)
+
+--- a/vorbis.c
++++ b/vorbis.c
+@@ -56,7 +56,7 @@
+ #define PROGRESS_FORMAT_SIZE 8
+ #define MIN_FILENAME_SIZE 5
+ #define MIN_MIDDLE_TRUNCATE_SIZE 20
+-#define TEMP_NAME "vorbisgain.tmpXXXXXX"
++#define TEMP_EXT ".vgain.tmpXXXXXX"
+ 
+ 
+ /**
+@@ -684,11 +684,10 @@
+         }
+     }
+ 
+-    /* Make sure temp is in same folder as file. And yes, the malloc is larger
+-     * than necessary (and not always needed). Lets keep it simple though (at
+-     * the expense of a few bytes)...
++    /* Construct a temporary file name by appending TEMP_EXT to
++     * the name of the file being modified.
+      */
+-    temp_name = malloc(strlen(filename) + sizeof(TEMP_NAME));
++    temp_name = malloc(strlen(filename)*sizeof(char) + sizeof(TEMP_EXT));
+ 
+     if (temp_name == NULL)
+     {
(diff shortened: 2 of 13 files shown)


CODEOWNERS
```diff
@@ -16,3 +16,4 @@
 /mac/              @enzo1982 @Eonfge
 /pygtk/            @Eonfge
 /gzdoom/           @Eonfge
+/vobisgain/        @Eonfge
```
hfiguiere, line 19:
```suggestion
/vorbisgain/        @Eonfge
```

Eonfge:
@enzo1982 Do you use vorbisgain in your application? I spotted some possible security vulnerabilities with the package as I use it with Flacon, and it might be best to keep the package patched where possible.

vorbisgain/vorbisgain_0.37-2.json
```diff
@@ -0,0 +1,51 @@
...
+    "buildsystem": "autotools",
+    "sources": [
+        {
+            "type": "archive",
+            "url": "http://deb.debian.org/debian/pool/main/v/vorbisgain/vorbisgain_0.37.orig.tar.gz",
+            "sha256": "dd6db051cad972bcac25d47b4a9e40e217bb548a1f16328eddbb4e66613530ec"
+        },
+        {
+            "type": "patch",
+            "path": "0001-temp_files.patch"
```
hfiguiere, line 12: it's not blcking but I think that using `paths` here with an array would be clearer and shorter. (applied in order)

hfiguiere (changes requested):

enzo1982:
> @enzo1982 Do you use vorbisgain in your application?

Thanks for the heads-up, but no, I don't use it.

vorbisgain/vorbisgain_0.37-2.json
Eonfge: :open_mouth: That is so much easier! I will do it right away!

Eonfge:
It's a real shame that these patches don't make it upstream, and as I mentioned in the readme, there are quite some distributions that don't have these patches, or that don't even ship the latest version. Goes to show how important sandboxing is, because it's easy for some common libraries to fall behind, even in existing distributions.

As for why I picked the Debian packages... they seem more complete... Fedora has its own set of patches for an older version:
https://src.fedoraproject.org/rpms/vorbisgain/tree/rawhide

vorbisgain/0001-temp_files.patch
```diff
@@ -0,0 +1,75 @@
...
+     }
+ 
+-    /* Make sure temp is in same folder as file. And yes, the malloc is larger
+-     * than necessary (and not always needed). Lets keep it simple though (at
+-     * the expense of a few bytes)...
++    /* Construct a temporary file name by appending TEMP_EXT to
++     * the name of the file being modified.
+      */
+-    temp_name = malloc(strlen(filename) + sizeof(TEMP_NAME));
++    temp_name = malloc(strlen(filename)*sizeof(char) + sizeof(TEMP_EXT));
```
TingPing, line 34: I'm pretty sure this patch isn't even correct. It needs `+ 1` for trailing `\0`.

vorbisgain/0003-manpage.patch
```diff
@@ -0,0 +1,248 @@
+Description: Man page says --silent, should be --skip
```
TingPing, line 1: This patch does nothing for flatpaks.

vorbisgain/0006-manpage_hyphens.patch
```diff
@@ -0,0 +1,56 @@
+Description: Resolved hyphen-used-as-minus-sign lintian complaints.
```
TingPing, line 1: Also not a useful patch.

vorbisgain/0009-hardening.patch
```diff
@@ -0,0 +1,13 @@
...
+Forwarded: no
+--- a/misc.c
++++ b/misc.c
+@@ -56,7 +56,7 @@
+     vfprintf(stderr, message, args);
+     va_end(args);
+ 
+-    fprintf(stderr, strerror(err_num));
++    fputs(strerror(err_num), stderr);
+     fprintf(stderr, "\n");
```
TingPing, line 11: I know you didn't write this but these patches are pretty lazy and low quality IMO.

This could be `fprintf("%s\n", strerror(err_num));`

vorbisgain/0008-manpage_recursion_mistake.patch
```diff
@@ -0,0 +1,23 @@
+Description: The manpage shows methods that don't work. They have been replaced.
```
TingPing, line 1: Another docs only patch.

vorbisgain/0001-temp_files.patch
TingPing: Oh, nevermind its actually correct as `TEMP_EXT` includes it.

I'd still call this low quality but oh well.

TingPing:
Its fine to me. I'll leave it up to you if the manpage fixes stay or go. Just seems necessary to me.

Eonfge:
@TingPing I decided to include the man-page patches from completeness sake. now I can just take the Debian patches and apply those, without thinking to much about it. In the end it's about ensuring easy maintenance now and in the future.

PR #174, opened 2022-02-02
Update libusb to 1.0.25

A6GibKm:
Tested that it builds with https://github.com/flathub/cc.arduino.arduinoide.


BigmenPixel0:
org.getmonero.Monero builds with this.

PR #175, opened 2022-02-04
Add pyqt5

catsout:
## pyqt5.yml

- This module will build and install two python packages to `${FLATPAK_DEST}`:  

```text
PyQt5-sip
PyQt5
```

- Support setting specific compile options  

  Like `--enable|disable QtGui`  
  Here are other [modules](https://doc.qt.io/qtforpython/modules.html) and other [options](https://www.riverbankcomputing.com/static/Docs/PyQt5/installation.html#building-and-installing-from-source)

  `confirm-license` and `no-dbus-python` are hard-coded

- This module will create `${FLATPAK_DEST}/build-pyqt5` and delete this folder during cleanup phase

### example

```yaml
app-id: org.flathub.app
runtime: org.kde.Platform
runtime-version: '5.15-21.08'
sdk: org.kde.Sdk
#...
# only global build-options can pass env to modules
# so put the opts here
build-options:
  env:
    PYQT5_OPTS: >-
      --no-designer-plugin
      --no-qml-plugin
      --no-tools
      --enable QtCore
      --enable QtGui
      --enable QtNetwork
      --enable QtWidgets
modules:
  - shared-modules/pyqt5/pyqt5.yml
  
  - name: app
    #...
```

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -17,3 +17,4 @@
 /pygtk/            @Eonfge
 /gzdoom/           @Eonfge
 /vorbisgain/       @Eonfge
+/pyqt5/            @catsout
--- a/pyqt5/README.md
+++ b/pyqt5/README.md
@@ -1,0 +1,49 @@
+## pyqt5.yml
+
+- This module will build and install two python packages to `${FLATPAK_DEST}`:  
+
+```text
+PyQt5-sip
+PyQt5
+```
+
+- Support setting specific compile options  
+
+  Like `--enable|disable QtGui`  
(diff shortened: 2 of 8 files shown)


catsout:
Tested on https://github.com/flathub/io.github.ciromattia.kcc

hfiguiere:
somebody needs to be added to CODEOWNERS

tinywrkb:
@catsout I think I'll submit a PyQtWebEngine base app in a week or so.  
I prefer a base app over a shared module, as it will:

* Shorter app's build run, saving time and CPU cycles for local and CI builds.
* Be automatically updated by Flathub's bot.

I can construct it in a way that if an app maintainer don't want to wait for a base app update from the repo, maybe in order to get a security fix faster, then the PyQtWebEngine or PyQt manifest from the base app's source can be used directly in the app's packaging.

Issue #178, opened 2022-02-07
Add clutter and clutter-gtk


A6GibKm:
Clutter and clutter-gtk are deprecated and are not going to be part of the GNOME 42 runtimes and some apps still rely on it, it could be a good idea to have it here as a low maintenance module.

For reference, this is how it looks on one of my projects.

```json
        {
            "name": "cogl",
            "config-opts": [
                "--disable-cogl-gst",
                "--enable-xlib-egl-platform",
                "--enable-wayland-egl-platform"
            ],
            "sources": [
                {
                    "type": "git",
                    "url": "https://gitlab.gnome.org/GNOME/cogl.git",
                    "branch": "cogl-1.22"
                }
            ]
        },
        {
            "name": "mtdev",
            "buildsystem": "autotools",
            "sources": [
                {
                    "type": "git",
                    "url": "https://github.com/rydberg/mtdev.git"
                }
            ]
        },
        {
            "name": "libinput",
            "buildsystem": "meson",
            "config-opts": [
                "-Dlibwacom=false",
                "-Dtests=false",
                "-Ddocumentation=false"
            ],
            "sources": [
                {
                    "type": "git",
                    "url": "https://gitlab.freedesktop.org/libinput/libinput.git"
                }
            ]
        },
        {
            "name": "clutter",
            "buildsystem": "meson",
            "config-opts": [
                "-Ddocumentation=false",
                "-Dbackends=x11,wayland"
            ],
            "sources": [
                {
                    "type": "git",
                    "url": "https://gitlab.gnome.org/GNOME/clutter.git"
                }
            ]
        },
        {
            "name": "clutter-gtk",
            "buildsystem": "autotools",
            "sources": [
                {
                    "type": "git",
                    "url": "https://gitlab.gnome.org/GNOME/clutter-gtk.git"
                }
            ]
        },
```

See https://gitlab.gnome.org/GNOME/gnome-nibbles/-/merge_requests/31.

hfiguiere:
submit a PR, set yourself in the owners to maintain it, and make sure you don't use bare URL or branches for the sources.

TingPing:
Which apps actually use it?

Totem will be porting away.

A6GibKm:
From gnome at least https://gitlab.gnome.org/GNOME/Initiatives/-/issues/31 use them. There are probably more apps outside of gnome-core and gnome-games.

PR #183, opened 2022-03-24
Add clutter module

A6GibKm:
GNOME 42 removed clutter and clutter-gtk from the SDK.

Fixes: https://github.com/flathub/shared-modules/issues/178

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -2,6 +2,7 @@
 *                  @flathub/reviewers
 
 /dbus-glib/        @TingPing
+/clutter/          @A6GibKm
 /gtk2/             @TingPing
 /gudev/            @Erick555
 /intltool/         @TingPing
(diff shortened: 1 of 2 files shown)


A6GibKm:
Given that the runtime just dropped it, it is very hard to eye ball how many apps actually need clutter. I tested and this builds just fine on top of the GNOME 42 runtime.

A6GibKm:
@ebassi do you think the config-opts used are sane?

ebassi:
The configuration options look good. You may want to add clutter-gst to the module, since people used it with clutter-gtk most of the time.

A6GibKm:
> The configuration options look good. You may want to add clutter-gst to the module, since people used it with clutter-gtk most of the time.

Thanks, I added clutter-gst.

PR #184, opened 2022-03-26
Fix checksum for ladspa

vchernin:
Caused some local build failures. This should not be merged without being certain this is correct, there is no recent changelog upstream which is a bit concerning:
https://www.ladspa.org/ladspa_sdk/changes.html

### DIFF
--- a/linux-audio/ladspa.json
+++ b/linux-audio/ladspa.json
@@ -11,7 +11,7 @@
         {
             "type": "archive",
             "url": "http://www.ladspa.org/download/ladspa_sdk_1.17.tgz",
-            "sha256": "d9d596171d93f9c226fcdb7e27c6f917422ac487efe2c05e0a18094df4268061"
+            "sha256": "27d24f279e4b81bd17ecbdcc38e4c42991bb388826c0b200067ce0eb59d3da5b"
         }
     ]
 }


hfiguiere:
I didn't see this PR. I just pushed a fix, I checked the archive content from the previous out. Identical, possibly a repack.

rncbc:
apparently this PR is still needed to correct checksum:
https://www.ladspa.org/download/ladspa_sdk_1.17.tgz 
on this date verified sha256sum: 27d24f279e4b81bd17ecbdcc38e4c42991bb388826c0b200067ce0eb59d3da5b

vchernin:
The same fix was commited to main: https://github.com/flathub/shared-modules/commit/81b7b525dbd79258a7e9fb3df57461977806fe69

rncbc:
and it failed again:
https://buildbot.flathub.org/#/builders/12/builds/11949
same reason (wrong sha256)

vchernin:
The shared modules commit is from 6 months ago: https://github.com/flathub/org.rncbc.qtractor/commit/1cd08bef9ec5464532f18ebd607088aaa2366aee

rncbc:
interesting, na may you tell how to `git submodule update` on the https://github.com/flathub/org.rncbc.qtractor ?

vchernin:
yeah, I can't think of a better way

rncbc:
ok, thanks, so be it

hfiguiere:
`cd shared-modules && git pull -r origin master`
And and then `git commit shared-modules -m "Update shared-modules"`

Also it's often a good idea to wait for the build to run on a PR. It's automatic.

PR #186, opened 2022-03-28
add boost

kevinsmia1939:
There are many package that use boost, so I think it is a good idea to put in share-modules.
I can be the maintainer.

### DIFF
--- a/boost/boost.json
+++ b/boost/boost.json
@@ -1,0 +1,21 @@
+{
+        "name": "boost",
+        "buildsystem": "simple",
+        "build-commands": [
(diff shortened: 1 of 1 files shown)


hfiguiere:
And most package pick and choose which module they want, and a lot of them just use headers.

Not sure this is the best usage of shared-modules.

TingPing:
Yes boost is a poor fit. It has many options that are often project specific. Out of curiosity how large do all modules end up being?

kevinsmia1939:
I see, boost might not be the best then.

PR #188, opened 2022-04-16
Use cflags and not env CFLAGS

hfiguiere:


TingPing:
LGTM whenever this is ready for merge.

PR #191, opened 2022-05-12
Added PhysicsFS

Mailaender:
https://github.com/search?q=org%3Aflathub+physfs&type=code and everyone is using the same version.

### DIFF
--- a/physfs/physfs.json
+++ b/physfs/physfs.json
@@ -1,0 +1,18 @@
+{
+    "name": "PhysicsFS",
+    "buildsystem": "cmake-ninja",
+    "config-opts": [ "-DPHYSFS_BUILD_TEST=OFF", "-DPHYSFS_BUILD_STATIC=OFF" ],
+    "cleanup": [
+        "/include",
+        "/lib/pkgconfig",
+        "/bin"
+    ],
+    "sources": [
+        {
+            "type": "archive",
+            "url": "https://icculus.org/physfs/downloads/physfs-3.0.2.tar.bz2",
+            "sha256": "304df76206d633df5360e738b138c94e82ccf086e50ba84f456d3f8432f9f863",
+            "size": 194888
+        }
+    ]
+}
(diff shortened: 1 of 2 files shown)


physfs/physfs.json
```diff
@@ -0,0 +1,24 @@
+{
+    "name": "PhysicsFS",
+    "buildsystem": "cmake-ninja",
+    "config-opts": [ "-DPHYSFS_BUILD_TEST=OFF" ],
+    "cleanup": [
+        "/lib/*.a",
```
Erick555, line 6: Would just `*.a` work?

physfs/physfs.json
```diff
@@ -0,0 +1,24 @@
...
+        "/lib/pkgconfig",
+        "/bin"
+    ],
+    "sources": [
+        {
+            "type": "archive",
+            "url": "https://icculus.org/physfs/downloads/physfs-3.0.2.tar.bz2",
+            "sha256": "304df76206d633df5360e738b138c94e82ccf086e50ba84f456d3f8432f9f863",
+            "size": 194888,
+            "x-checker-data": {
```
hfiguiere, line 17: I don't think the x-checker-data works with shared-modules.

hfiguiere:
> https://github.com/search?q=org%3Aflathub+physfs&type=code and everyone is using the same version.

and it's really trivial do build. Doesn't require special patches or hackery.

physfs/physfs.json
hfiguiere: also the config `-DPHYSFS_BUILD_STATIC=OFF` would also do it (better)

Mailaender:
Updated.

TingPing:
Need somebody added to `CODEOWNERS` to commit to updating it.

Mailaender:
Added myself.

barthalion:
Please make sure to submit pull requests to repos which could use it.

Issue #193, opened 2022-06-13
clutter: Disable gtk-doc generation


wjt:
I built a test app that just builds `shared-modules/clutter/clutter.json`. It's 22 MB of which 8 MB is API documentation.

```
$ (cd `flatpak info -l app/uk.me.wjt.Test/x86_64/master`; ( du -h --max-depth=2 files | sort -h | tail -n5 ))
3.3M	files/lib
6.4M	files/share/gir-1.0
8.1M	files/share/gtk-doc
15M	files/share
21M	files
```

I tried writing the obvious patch but it had no effect.

```
commit 0ada6db8c75ee0193f3d937e01e297efee84df1a (HEAD -> clutter-cleanup)
Author: Will Thompson <will@willthompson.co.uk>
Date:   Mon Jun 13 14:46:46 2022 +0100

    clutter: Disable gtk-doc in all modules
    
    This makes for a faster build and smaller apps.

diff --git a/clutter/clutter.json b/clutter/clutter.json
index cb6baaa..4a964f8 100644
--- a/clutter/clutter.json
+++ b/clutter/clutter.json
@@ -12,6 +12,7 @@
             "name": "cogl",
             "config-opts": [
                 "--disable-cogl-gst",
+                "--disable-gtk-doc",
                 "--enable-xlib-egl-platform",
                 "--enable-wayland-egl-platform"
             ],
@@ -26,6 +27,7 @@
         {
             "name": "clutter",
             "config-opts": [
+                "--disable-gtk-doc",
                 "--enable-egl-backend",
                 "--enable-wayland-backend"
             ],
@@ -39,6 +41,9 @@
         },
         {
             "name": "clutter-gst",
+            "config-opts": [
+                "--disable-gtk-doc"
+            ],
             "sources": [
                 {
                     "type": "archive",
```

So I have given up.

Erick555:
Maybe try `--enable-docs=no` . I see it also have meson buildsystem but I don't know if it's usable. If everything else fails then you may just cleanup those docs.

PR #194, opened 2022-06-13
clutter: Remove docs from build

A6GibKm:
Fixes: https://github.com/flathub/shared-modules/issues/193

### DIFF
--- a/clutter/clutter.json
+++ b/clutter/clutter.json
@@ -1,5 +1,8 @@
 {
     "name": "clutter-gtk",
+    "cleanup": [
+        "/share/gtk-doc"
+    ],
     "sources": [
         {
             "type": "archive",
@@ -12,8 +15,13 @@
             "name": "cogl",
             "config-opts": [
                 "--disable-cogl-gst",
+                "--disable-gtk-doc",
                 "--enable-xlib-egl-platform",
                 "--enable-wayland-egl-platform"
+            ],
+            "cleanup": [
+                "/share/gtk-doc",
+                "/share/cogl/examples-data"
             ],
             "sources": [
                 {
@@ -26,8 +34,12 @@
         {
             "name": "clutter",
             "config-opts": [
+                "--disable-gtk-doc",
                 "--enable-egl-backend",
                 "--enable-wayland-backend"
+            ],
+            "cleanup": [
+                "/share/gtk-doc"
             ],
             "sources": [
                 {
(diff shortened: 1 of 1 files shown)


clutter/clutter.json
```diff
@@ -1,5 +1,8 @@
 {
     "name": "clutter-gtk",
+    "cleanup": [
+        "/share/gtk-doc"
```
wjt, line 4: This is insufficient, you need to clean up the submodules too.

wjt:
```
diff --git a/clutter/clutter.json b/clutter/clutter.json
index 6795c98..f7b9b04 100644
--- a/clutter/clutter.json
+++ b/clutter/clutter.json
@@ -13,6 +13,9 @@
     "modules": [
         {
             "name": "cogl",
+            "cleanup": [
+                "/share/gtk-doc"
+            ],
             "config-opts": [
                 "--disable-cogl-gst",
                 "--disable-gtk-doc",
@@ -29,6 +32,9 @@
         },
         {
             "name": "clutter",
+            "cleanup": [
+                "/share/gtk-doc"
+            ],
             "config-opts": [
                 "--disable-gtk-doc",
                 "--enable-egl-backend",
@@ -44,6 +50,9 @@
         },
         {
             "name": "clutter-gst",
+            "cleanup": [
+                "/share/gtk-doc"
+            ],
             "config-opts": [
                 "--disable-gtk-doc"
             ],
```

A6GibKm:
done

wjt:
Looks good. If you're feeling keen you could also add:

```patch
diff --git a/clutter/clutter.json b/clutter/clutter.json
index 0f0967a..a47b207 100644
--- a/clutter/clutter.json
+++ b/clutter/clutter.json
@@ -20,7 +20,8 @@
                 "--enable-wayland-egl-platform"
             ],
             "cleanup": [
-                "/share/gtk-doc"
+                "/share/gtk-doc",
+                "/share/cogl/examples-data"
             ],
             "sources": [
                 {
```

There is some other stuff that could be cleaned up, like some valgrind suppressions and all of `/include`, but this has knocked 8MB off. Of the remaining 13M the biggest piece is the gobject-introspection stuff, which the app might need.

A6GibKm:
Thanks, I removed a cleanup for the examples.

The introspection stuff is massive, I wish we had a way to pass arguments  like `--enable-instrospection=false` to shared modules.

hfiguiere:
> Thanks, I removed a cleanup for the examples.
> 
> The introspection stuff is massive, I wish we had a way to pass arguments like `--enable-instrospection=false` to shared modules.

`--disable-introspection` doesn't work?

A6GibKm:
> > Thanks, I removed a cleanup for the examples.
> > The introspection stuff is massive, I wish we had a way to pass arguments like `--enable-instrospection=false` to shared modules.
> 
> `--disable-introspection` doesn't work?

I don't see why not, but then we either have two provide two versions of the module or we left out apps that use introspection data.

PR #197, opened 2022-08-03
lua5.4: Fix json for liblua.so.5.4

ptitjano:


hfiguiere:
Are you sure you are not confusing soname and version?

ptitjano:
When I use current master on a flatpak application, I get this error: 

`Failed to open module: liblua.so.5.4: cannot open shared object file: No such file or directory`

I thought this change would fix the issue.

JBBgameich:
I can confirm this patch allows running Ultrastar Deluxe on top of this shared-module

hfiguiere:
thanks!

PR #199, opened 2022-08-04
fix(libusb): switch to .tar.bz2 tarball

Lctrs:
According to a txt shipped with the [release](https://github.com/libusb/libusb/releases/tag/v1.0.26) on their github, we should use the .tar.bz2 file instead of the .tar.gz.


BigmenPixel0 (approved):

A6GibKm:
lgtm, thanks.

PR #208, opened 2022-09-18
Update to fluidsynth 2.2.9

pedrolcl:
FluidSynth 2.2.9 has been released only a few days ago.

I've realized that I would like to add a configuration option like this one:

~~~
    "-DDEFAULT_SOUNDFONT=$HOME/.local/share/soundfonts/default.sf2"
~~~

To override the default hardcoded value in fluidsynth, which is: `/usr/share/soundfonts/default.sf2`.

But obviously resolving `$HOME` at runtime, not at configuration time. I know the environment variables `$XDG_DATA_DIRS`  and `$XDG_DATA_HOME`, but I have the same problem as with `$HOME`. 

It would be nice to be able to also provide soundfont flatpaks, leveraging a shared default location.

### DIFF
--- a/linux-audio/fluidsynth2-static.json
+++ b/linux-audio/fluidsynth2-static.json
@@ -20,8 +20,8 @@
   "sources": [
     {
       "type": "archive",
-      "url": "https://github.com/FluidSynth/fluidsynth/archive/v2.2.8.tar.gz",
-      "sha256": "7c29a5cb7a2755c8012d941d1335da7bda957bbb0a86b7c59215d26773bb51fe"
+      "url": "https://github.com/FluidSynth/fluidsynth/archive/v2.2.9.tar.gz",
+      "sha256": "bc62494ec2554fdcfc01512a2580f12fc1e1b01ce37a18b370dd7902af7a8159"
     }
   ]
 }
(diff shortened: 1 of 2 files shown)


hfiguiere:
this won't work for many reason. the first one being that `.config` isn't available by default.

Not that it could be `${FLATPAK_DEST}/share/soundfonts/default.sf2` but the file would need to be provided.

A lot of the "UNIX assumption" are broken with sandboxing.

pedrolcl:
> this won't work for many reason. the first one being that `.config` isn't available by default.

You probably mean `~/.local`, right?

> Not that it could be `${FLATPAK_DEST}/share/soundfonts/default.sf2` but the file would need to be provided.

The file name `default.sf2` is simply a placeholder, so the packager (or administrator, or user) could provide a real preferred soundfont file, symlinked with this name. It is only a hardcoded string inside the fluidsynth library. The goal is to find a valid SF2 even if the application using fluidsynth doesn't provide a file name.

hfiguiere:
yeah I meant `.local`. same problem though

pedrolcl:
Something like https://github.com/flathub/org.gustavoperedo.FontDownloader would also be nice, but for soundfonts instead of typographic fonts.

PR #209, opened 2022-09-19
Move all GZDoom dependencies in a single file

Eonfge:


Eonfge:
@hfiguiere Weird question perhaps, but do these Anitya checks work for shared modules? It helps me maintain updates, but I then know if I have to do it manually

PR #210, opened 2022-09-25
update FluidSynth to v2.3.0

pedrolcl:
This FluidSynth release is ABI compatible with v2.2.x but the build system has changed:
* LIB_SUFFIX option removed. Now using cmake module GNUInstallDirs.
* CMake exported targets, now supporting find_package(FluidSynth).


pedrolcl:
@hfiguiere ping!

PR #212, opened 2022-10-08
Add Fluidsynth 2.1.9 module.

### LINKED ISSUE #211: Fluidsynth 2.2.0 SOVERSION bump
Fluidsynth version 2.2.0 introduced some [breaking changes](https://www.fluidsynth.org/api/RecentChanges.html#NewIn2_2_0) and bumped Fluidsynth's SOVERSION from 2 to 3.
For backwards compatibility purposes, it might be a good idea to rename the current fluidsynth modules to `fluidsynth3.json` and so on and create a new `fluidsynth2.json` module that specifies version 2.1.9 (the latest fluidsynth version with SOVERSION 2)

The reason I'm opening this issue is because I'm trying to create a flatpak for a program but it depends on `libfluidsynth.so.2`, whereas the current `fluidsynth2.json` module in this repo builds and creates `libfluidsynth.so.3` instead.

Zumorica:
Fixes #211


linux-audio/fluidsynth2-static.json
```diff
@@ -2,16 +2,18 @@
   "name": "fluidsynth",
   "buildsystem": "cmake-ninja",
   "build-options": {
-    "cflags": "-fPIC",
-    "cxxflags": "-fPIC"
+    "env": {
```
Erick555, line 5: This overrides default CFLAGS/CXXFLAGS variables. Don't do this. How this change is even related to this PR?

Erick555:
I wonder if it will be clear to people that fluidsynth3 refers to 2.3.

hfiguiere:
sorry, no. See my other comment on #211.

linux-audio/fluidsynth2-static.json
Zumorica: I literally just grabbed the file from commit https://github.com/flathub/shared-modules/commit/a79072c2f6cd6c5b1a6cb8ccec649cfb87608926 and replaced the fluidsynth version, I didn't make any more changes.

PR #213, opened 2022-11-01
lua: Actually clean up manpages

wjt:
For reasons unknown, Lua does not install its manpages to $PREFIX/share/man, but to $PREFIX/man.

The lua5.1 module already listed "/man" rather than "/share/man" in its cleanups.


wjt:
Draft while I test this…

Issue #217, opened 2022-12-08
SDL2 submodule, with libdecor


orowith2os:
SDL2 seems to have gotten one of its final releases, and the SDL team is doing some major cleanup for SDL3.

I believe it would be beneficial to add an SDL2 submodule that functions similarly to the original SDL submodule(s).

Is there anything I should add alongside a basic SDL2 and libdecor submodule? How it should be structured might be concerning too, should I have them both be in the same folder (GIT_ROOT/SDL2/{sdl2.json, libdecor.json}. sdl2.json pulls in libdecor.json during build) or separate submodules?

TingPing:
I would have "sdl2.json" implicitly pull in "libdecor.json" at all times so users of this never need to think about it.

They can be in their own folders so people can individually use libdecor if they wish.

Make sure you update `CODEOWNERS` with somebody to maintain these modules.

orowith2os:
@TingPing So, in a separate MR, add sdl2 and then link back to this libdecor submodule using ../libdecor/libdecor-ver.json?

TingPing:
Yup.

orowith2os:
> @orowith2os Actually I'm a bit confused, `sdl2` is inside the freedesktop-sdk. Have you tried getting `libdecor` added there as well?

It was discussed a while back, but denied. And just building libdecor inside of flatpak doesn't allow the runtime SDL2 to use that either, so that's my main motivation.

One moment, I'll see if I can drag up the exact issue...

TingPing:
OK. It is fine to keep here for a bit. The ultimate goal next year should be a push to get it into fdo though IMO.

orowith2os:
I would like to do that, yes, though at that point, I'd prefer it that the people there manage it; I don't like using gitlab for interacting with other users.

orowith2os:
Opting to exclude linking to libdecor in the SDL2 manifest, otherwise managing it and other, alternative, SDL2 manifests might serve to be a pain.

PR #218, opened 2022-12-08
Add libdecor 0.1.1

orowith2os:
For libdecor 0.2.0 (or the next release, whatever), it'll need to have GTK as a build dep for the GTK plugin. This could be done in another module, and then pulled in by libdecor. I'll handle it when the time comes.

This is intended to be used alongside an SDL2 shared module, but could be useful for https://github.com/flathub/org.blender.Blender/pull/126

### DIFF
--- a/libdecor/libdecor-0.1.1.json
+++ b/libdecor/libdecor-0.1.1.json
@@ -6,9 +6,9 @@
   ],
   "sources": [
     {
-      "type": "git",
-      "url": "https://gitlab.freedesktop.org/libdecor/libdecor.git",
-      "tag": "0.1.1"
+      "type": "archive",
+      "url": "https://gitlab.freedesktop.org/libdecor/libdecor/-/archive/0.1.1/libdecor-0.1.1.tar.gz",
+      "sha256": "82adece5baeb6194292b0d1a91b4b3d10da41115f352a5e6c5844b20b88a0512"
     }
   ]
 }
--- a/libdecor/libdecor-0.1.1.json
+++ b/libdecor/libdecor-0.1.1.json
@@ -10,5 +10,8 @@
       "url": "https://gitlab.freedesktop.org/libdecor/libdecor/-/archive/0.1.1/libdecor-0.1.1.tar.gz",
       "sha256": "82adece5baeb6194292b0d1a91b4b3d10da41115f352a5e6c5844b20b88a0512"
     }
-  ]
+  ],
+  "cleanup": [
+    "/include",
+    "/lib/pkgconfig"
 }
--- a/libdecor/libdecor-0.1.1.json
+++ b/libdecor/libdecor-0.1.1.json
@@ -14,4 +14,5 @@
   "cleanup": [
     "/include",
     "/lib/pkgconfig"
+  ]
 }
--- a/libdecor/libdecor-0.1.1.json
+++ b/libdecor/libdecor-0.1.1.json
@@ -1,0 +1,14 @@
+{
+  "name": "libdecor",
+  "buildsystem": "meson",
+  "config-opts": [
+    "-Ddemo=false"
+  ],
+  "sources": [
+    {
+      "type": "git",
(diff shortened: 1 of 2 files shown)


orowith2os:
Partially closes https://github.com/flathub/shared-modules/issues/217

orowith2os:
the GTK plugin is only in -git, and I thought GTK was only in the GNOME platform. If it's in the KDE and fd.o platforms, then when the time comes to add a new libdecor release, it should be fine.

hfiguiere:
gtk3 has always been in the freedesktop-sdk.

orowith2os:
That's good to know, I haven't really dug around the sdk unless it's for something I *really* need. 

Aside from that, what about making this use YAML instead? Does Flatpak handle that fine, or is JSON preferred for shared modules?

Pixelnarium:
While GTK3 is in the runtime I don't know if you need the GNOME runtime to get a proper looking titlebar.

TingPing:
> Aside from that, what about making this use YAML instead? Does Flatpak handle that fine, or is JSON preferred for shared modules?

I would like to keep it as JSON for consistency. Though both file types work the same way.

orowith2os:
> While GTK3 is in the runtime I don't know if you need the GNOME runtime to get a proper looking titlebar.

Libdecor only specifies GTK3 for the GTK CSD, so I don't think there should be any issues on that front. Only libdecor or the app handling setting the CSD to the GTK pugin.

Again, I'll handle it when the time comes, libdecor still has some issues with plugins that (imo) should be cleared up first, like auto setting the appropriate plugin.

Pixelnarium:
> There is no difference in GTK3 in fdo and GTK3 in GNOME.

I am thinking more about the permissions and portals to access some host folders to get proper styling information about the current theme in use. But sure my knowledge around this topic is rather small.

TingPing:
> I am thinking more about the permissions and portals to access some host folders to get proper styling information about the current theme in use.

All apps have access to the portal, its not an opt-in permission. `xdg-desktop-portal-gtk` does need to be installed on the host though.

libdecor/libdecor-0.1.1.json
```diff
@@ -0,0 +1,14 @@
+{
+  "name": "libdecor",
+  "buildsystem": "meson",
```
TingPing, line 3: Missing `"cleanup"` for all development files installed.

libdecor/libdecor-0.1.1.json
```diff
@@ -0,0 +1,14 @@
...
+  "config-opts": [
+    "-Ddemo=false"
+  ],
+  "sources": [
+    {
+      "type": "git",
+      "url": "https://gitlab.freedesktop.org/libdecor/libdecor.git",
```
TingPing, line 10: Could you use release tarballs, they will speed up build times for every project using this.

https://gitlab.freedesktop.org/libdecor/libdecor/uploads/ee5ef0f2c3a4743e8501a855d61cb397/libdecor-0.1.1.tar.xz

TingPing (commented):
Please add yourself to `CODEOWNERS`. Somebody needs to be responsible for maintaining this.

libdecor/libdecor-0.1.1.json
orowith2os: I believe I've added the source properly now

libdecor/libdecor-0.1.1.json
orowith2os: Ah, I don't think I know the specific devel files I'll need to clean up >>

Would I need to go into the flatpak environment and see what I need to get rid of?

orowith2os: Okay, I *think* I've found the stuff I need to clean up? Would adding these be fine? https://github.com/flathub/org.nongnu.enigma/blob/62d92b83efec2c3c1bc3fd217f53e204fb977993/org.nongnu.enigma.yml#L42

PR #220, opened 2022-12-09
Add READMEs to SDL1 and libdecor

orowith2os:
IMO, other shared modules should follow this practice to be more verbose on how they're intended to be used. This is particularly useful in the case of something like libdecor, where using a specific runtime might be a concern (it was for me when I was trying to get GTK CSDs to work).

### DIFF
--- a/libdecor/README.md
+++ b/libdecor/README.md
@@ -2,5 +2,5 @@ This is a shared module for libdecoration, used by several applications and util
 
 This will contain the latest versioned release of libdecor; if newer (devel) versions are desired, please manually add it to your flatpak manifest.
 
-The freedesktop runtime already bundles GTK3, so you don't need to do any more work to get native-looking CSDs on GNOME, at the very least.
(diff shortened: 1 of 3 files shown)


libdecor/README.md
```diff
@@ -0,0 +1,6 @@
+This is a shared module for libdecoration, used by several applications and utilities to get window decorations on Wayland compositors that might not implement server-side decorations.
+
+This will contain the latest versioned release of libdecor; if newer (devel) versions are desired, please manually add it to your flatpak manifest.
+
+The freedesktop runtime already bundles GTK3, so you don't need to do any more work to get native-looking CSDs on GNOME, at the very least.
```
TingPing, line 5: I thought this wasn't true until the gtk plugin is in a stable release?

orowith2os: Oh, yes, that is true. I'll update that section.

TingPing (approved):
:+1:

PR #221, opened 2022-12-09
Add GameMode shared module

orowith2os:
Particularly useful for a number of apps on Flathub, such as Lutris, Steam, Bottles, and WarThunder.

I've separated the libraries and script so that apps can add the bare minimum of what they need.


gamemode/gamemoderun.json
```diff
@@ -0,0 +1,15 @@
+{
+  "name": "gamemoderun",
+  "buildsystem": "simple",
+  "build-commands": [
+    "mv data/gamemoderun /app/bin/gamemoderun",
```
TingPing, line 5: This should be built from source.

TingPing: Oh its just a shell script.

You could simplify this by only downloading a single file.

orowith2os: It doesn't need to be, as it's a shell script. I tried using the `install` command for this, but couldn't quite figure it out, so I fell back to `mv` and `chmod`.

gamemode/README.md
```diff
@@ -0,0 +1,5 @@
+This is a submodule for FeralInteractive's GameMode.
+
+To use this, add the manifests listed here; `libgamemode*.json` will install the shared libraries, `gamemoderun.json` will install `gamemoderun` to `/app/bin`.
+
+You will need to add both manifests separately.
```
TingPing, line 5: What's the harm in `libgamemode` always installing `gamemoderun`?

orowith2os: with libgamemode, apparently the meson buildsystem doesn't install gamemoderun to the prefix, no matter the config-opts. I can make an issue in upstream gamemode about it, or see if I can find something about why it's not done.

gamemode/libgamemode-1.7.json
```diff
@@ -0,0 +1,17 @@
...
+    "-Dwith-sd-bus-provider=no-daemon" 
+  ],
+  "sources": [
+    {
+      "type": "archive",
+      "url": "https://github.com/FeralInteractive/gamemode/archive/refs/tags/1.7.tar.gz",
+      "sha256": "8a1718c657cb1d3c8269edfd397faf3cb41fccd885198ef0225bf8f82217c42f"
+    }
+  ],
+  "cleanup": [ "/include", "/lib/pkgconfig", "/lib/libgamemodeauto.a" ]
```
TingPing, line 16: Try `-Ddefault_library=shared` and see if it still builds a static lib.

gamemode/README.md
orowith2os: I'll merge the two manifests, it's not a large deal to add a separate module for gamemoderun anyways.

gamemode/libgamemode-1.7.json
TingPing: Oh I see it hardcodes builting both libraries.

Change this to a `"post-install"` and `rm` the shared library. We don't want projects accidentally linking to it.

gamemode/README.md
TingPing: You can add a custom `"post-install"` command to install the script also.

TingPing: Looking at their meson files they correctly install it into the right prefix..

TingPing: It was fixed in https://github.com/FeralInteractive/gamemode/commit/85d75c303c61d75602a8415716aa9d4befbf6f8a

orowith2os: > Looking at their meson files they correctly install it into the right prefix..

Interesting, it didn't do that when I built it without the config-opts. Is there anything I need to do to get it to install?

TingPing: It is behind the `if sd_bus_provider != 'no-daemon'` option.

TingPing: I think their logic is just wrong. A manual install seems fine.

orowith2os:
Simplified it a bit, so now it handles everything in one manifest.

gamemode/gamemode-1.7.json
```diff
@@ -13,5 +13,11 @@
       "sha256": "8a1718c657cb1d3c8269edfd397faf3cb41fccd885198ef0225bf8f82217c42f"
     }
   ],
-  "cleanup": [ "/include", "/lib/pkgconfig", "/lib/libgamemodeauto.a" ]
+  "cleanup": [ "/include", "/lib/pkgconfig" ],
+  "post-install": [
+    "mkdir -p /app/bin/",
+    "mv ../data/gamemoderun /app/bin/gamemoderun",
+    "chmod +x /app/bin/gamemoderun",
```
TingPing, line 20:
```suggestion
    "install -t /app/bin -Dm755 ../data/gamemoderun",
```

This probably works.

gamemode/gamemode-1.7.json
```diff
@@ -0,0 +1,23 @@
...
+      "type": "archive",
+      "url": "https://github.com/FeralInteractive/gamemode/archive/refs/tags/1.7.tar.gz",
+      "sha256": "8a1718c657cb1d3c8269edfd397faf3cb41fccd885198ef0225bf8f82217c42f"
+    }
+  ],
+  "cleanup": [ "/include", "/lib/pkgconfig" ],
+  "post-install": [
+    "mkdir -p /app/bin/",
+    "mv ../data/gamemoderun /app/bin/gamemoderun",
+    "chmod +x /app/bin/gamemoderun",
```
TingPing, line 20:
```suggestion
    "install -t /app/bin -Dm755 ../data/gamemoderun",
```

This probably works.

A6GibKm:
By the way, for lutris we also do a 32bit build (is this needed like at all?), this would make this shared module useless for us.

orowith2os:
> By the way, for lutris we also do a 32bit build (is this needed like at all?), can that be done with this shared-module?

I don't *believe* a 32-bit build is needed, but I could look into doing a 32-bit one as well. Though, it might be a bit weird, as apparently Lutris adds an ld flag and libdir.... TingPing might be able to give me some tips on that, if those are needed.

<https://github.com/flathub/net.lutris.Lutris/blob/beta/net.lutris.Lutris.yml#L70>
<https://github.com/flathub/net.lutris.Lutris/blob/beta/net.lutris.Lutris.yml#L283>

orowith2os:
Could I have some tips? I don't know how I would go about adding the 32-bit variant, especially since it seems to need some modifications to the ldflags and whatnot.

Doomsdayrs:
@orowith2os I sent a PR to your repo, Let me get to work

Doomsdayrs:
> Could I have some tips? I don't know how I would go about adding the 32-bit variant, especially since it seems to need some modifications to the ldflags and whatnot.

Lutris seems to do it via some flags, which seems very application specific

orowith2os:
Update comment: It seems like I can add 32-bit builds fairly easily, but it might be safer to separate them into their own manifest? 

I'll also stop versioning this when I get to updating it, for the same reasons I'll stop versioning the SDL2 module: there isn't much reason to it, since it's more or less compatible across versions.

RobLoach:
Was looking for this one. Any updates needed?

gamemode/gamemode-1.7.json
```diff
@@ -0,0 +1,21 @@
...
+    {
+      "type": "archive",
+      "url": "https://github.com/FeralInteractive/gamemode/archive/refs/tags/1.7.tar.gz",
+      "sha256": "8a1718c657cb1d3c8269edfd397faf3cb41fccd885198ef0225bf8f82217c42f"
+    }
+  ],
+  "cleanup": [ "/include", "/lib/pkgconfig" ],
+  "post-install": [
+    "install -t /app/bin -Dm755 ../data/gamemoderun",
+    "rm /app/lib/libgamemodeauto.a"
```
ranisalt, line 19: Add `/lib/*.a` to `cleanup` instead

orowith2os:
> Any updates needed?

I need to manage 32-bit builds somehow, and I'm not sure on how to do that. I also need to fix up the cleanup stuff :P

I'll get to this when I have a computer set up

gamemode/gamemode-1.7.json
```diff
@@ -0,0 +1,21 @@
...
+    "-Dwith-sd-bus-provider=no-daemon"
+  ],
+  "sources": [
+    {
+      "type": "archive",
+      "url": "https://github.com/FeralInteractive/gamemode/archive/refs/tags/1.7.tar.gz",
+      "sha256": "8a1718c657cb1d3c8269edfd397faf3cb41fccd885198ef0225bf8f82217c42f"
+    }
+  ],
+  "cleanup": [ "/include", "/lib/pkgconfig" ],
```
RobLoach, line 16:
``` suggestion
  "cleanup": [ 
    "/include",
    "/lib/pkgconfig",
    "/lib/*.a"
  ],
```

gamemode/gamemode-1.7.json
RobLoach, line 19:
``` suggestion
    "install -t /app/bin -Dm755 ../data/gamemoderun"
```

hfiguiere: Just add `*.a` like everywhere else.

orowith2os:
I'm still not sure how I should handle 32-bit builds, until that's figured out, this is blocked.

Doomsdayrs:
> I'm still not sure how I should handle 32-bit builds, until that's figured out, this is blocked.

Who can benefit from the 32bit build currently?

orowith2os:
> Who can benefit from the 32bit build currently?

Basically anybody Wine-based.

orowith2os:
I'll close this in favor of https://gitlab.com/freedesktop-sdk/freedesktop-sdk/-/issues/1641.

PR #224, opened 2022-12-13
libcanberra: Clean up random files

wjt:
This module leaves so much stuff lying around that is not needed at
runtime. Clean it all up so that apps don't have to.

### DIFF
--- a/libcanberra/libcanberra.json
+++ b/libcanberra/libcanberra.json
@@ -5,9 +5,15 @@
         "/bin",
         "/etc",
         "/include",
+        "/lib/gnome-settings-daemon-3.0",
+        "/lib/pkgconfig",
         "/libexec",
+        "/share/doc",
+        "/share/gdm",
+        "/share/gnome",
         "/share/gtk-doc",
-        "/share/man"
+        "/share/man",
+        "/share/vala"
     ],
     "config-opts": [
         "--disable-static",
--- a/libcanberra/libcanberra.json
+++ b/libcanberra/libcanberra.json
@@ -6,6 +6,7 @@
         "/etc",
         "/include",
         "/lib/gnome-settings-daemon-3.0",
+        "/lib/gtk-3.0",
         "/lib/pkgconfig",
         "/libexec",
         "/share/doc",


wjt:
OK, I'll change the message. But I think this is better anyway. libcanberra leaves so much junk around!

hadess:
> OK, I'll change the message. But I think this is better anyway. libcanberra leaves so much junk around!

I'm fine with it, but do you have a before and after list of files? I don't really have a test manifest at hand...

wjt:
```yaml
app-id: com.example.libcanberra
sdk: org.freedesktop.Sdk
runtime: org.freedesktop.Platform
runtime-version: '22.08'
command: /usr/bin/true
modules:
  - libcanberra.json
```

Here is the diff in the files without this change, and the files with this change:

```diff
--- /tmp/before.txt	2022-12-13 13:27:25.651256876 +0000
+++ /tmp/after.txt	2022-12-13 13:27:02.341100783 +0000
@@ -1,44 +1,22 @@
 .
 ./.ref
 ./manifest.json
 ./lib
-./lib/pkgconfig
-./lib/pkgconfig/libcanberra.pc
-./lib/pkgconfig/libcanberra-gtk3.pc
-./lib/gnome-settings-daemon-3.0
-./lib/gnome-settings-daemon-3.0/gtk-modules
-./lib/gnome-settings-daemon-3.0/gtk-modules/canberra-gtk-module.desktop
 ./lib/debug
 ./lib/libcanberra.so
 ./lib/libcanberra.so.0.2.5
 ./lib/libcanberra-0.30
 ./lib/libcanberra-0.30/libcanberra-multi.so
 ./lib/libcanberra-0.30/libcanberra-alsa.so
 ./lib/libcanberra-0.30/libcanberra-null.so
 ./lib/libcanberra-0.30/libcanberra-gstreamer.so
 ./lib/libcanberra-0.30/libcanberra-pulse.so
 ./lib/libcanberra.so.0
 ./lib/libcanberra-gtk3.so.0.1.9
 ./lib/gtk-3.0
 ./lib/gtk-3.0/modules
 ./lib/gtk-3.0/modules/libcanberra-gtk3-module.so
 ./lib/gtk-3.0/modules/libcanberra-gtk-module.so
 ./lib/libcanberra-gtk3.so
 ./lib/libcanberra-gtk3.so.0
 ./share
-./share/gnome
-./share/gnome/autostart
-./share/gnome/autostart/libcanberra-login-sound.desktop
-./share/gnome/shutdown
-./share/gnome/shutdown/libcanberra-logout-sound.sh
-./share/doc
-./share/doc/libcanberra
-./share/doc/libcanberra/README
-./share/gdm
-./share/gdm/autostart
-./share/gdm/autostart/LoginWindow
-./share/gdm/autostart/LoginWindow/libcanberra-ready-sound.desktop
-./share/vala
-./share/vala/vapi
-./share/vala/vapi/libcanberra-gtk.vapi
-./share/vala/vapi/libcanberra.vapi
```

hadess:
Should we also remove the GTK-related files? That might be a different issue though...

wjt:
Do you mean `canberra-gtk[3]-module`? (One is a symlink to the other.) I am not super familiar with how libcanberra works but I had assumed this was an important part of the library.

TingPing:
libcanberra is a standalone library on its own that applications can use to play sounds.

It also provides a GTK3 module that, on X11 only I believe, it can play event files on events like every time you click.

In practice I don't think anybody installs sound files for clicks in GTK. And with GTK4 and Wayland it does nothing anyway.

I think its reasonable to build it without that module.

wjt:
OK, I pushed an extra commit to delete that. (There's no configure flag to not build it.)

```diff
--- /tmp/after.txt	2022-12-14 14:25:59.944037233 +0000
+++ /tmp/after-2.txt	2022-12-14 14:26:09.732021389 +0000
@@ -1,22 +1,18 @@
 .
 ./.ref
 ./manifest.json
 ./lib
 ./lib/debug
 ./lib/libcanberra.so
 ./lib/libcanberra.so.0.2.5
 ./lib/libcanberra-0.30
 ./lib/libcanberra-0.30/libcanberra-multi.so
 ./lib/libcanberra-0.30/libcanberra-alsa.so
 ./lib/libcanberra-0.30/libcanberra-null.so
 ./lib/libcanberra-0.30/libcanberra-gstreamer.so
 ./lib/libcanberra-0.30/libcanberra-pulse.so
 ./lib/libcanberra.so.0
 ./lib/libcanberra-gtk3.so.0.1.9
-./lib/gtk-3.0
-./lib/gtk-3.0/modules
-./lib/gtk-3.0/modules/libcanberra-gtk3-module.so
-./lib/gtk-3.0/modules/libcanberra-gtk-module.so
 ./lib/libcanberra-gtk3.so
 ./lib/libcanberra-gtk3.so.0
 ./share
```

PR #228, opened 2023-01-01
Use $FLATPAK_DEST in post-install

hfiguiere:


hfiguiere:
Whithout this, it fails with any non standard prefix.

PR #229, opened 2023-01-02
Use ${FLATPAK_DEST} instead of hardcoding /app

o-kotb:
The path is different when building extensions, so it fails when it copies the license.


o-kotb:
https://github.com/flathub/shared-modules/pull/221

PR #231, opened 2023-01-07
Update SDL2 to 2.26.2

orowith2os:
https://github.com/libsdl-org/SDL/releases/tag/release-2.26.2


TingPing:
I think we should remove the version numbers from the files, just so its easier for projects to update.

orowith2os:
Fair, there isn't really much danger to automatically updating (dependabot) anyways. My hands are a bit full, will get onto it (if you haven't yet) when I can.

orowith2os:
Alright, removed the version numbers.

PR #232, opened 2023-02-17
Add libmpv

plata:


plata:
CC @espidev

libmpv/libmpv.json
```diff
@@ -0,0 +1,169 @@
...
+    "cleanup": [
+        "/include",
+        "/lib/pkgconfig",
+        "/share/man"
+    ],
+    "buildsystem": "simple",
+    "build-commands": [
+        "python3 waf configure --prefix=/app --enable-libmpv-shared --disable-cplayer --disable-build-date --disable-alsa",
+        "python3 waf build",
+        "python3 waf install"
```
hfiguiere, line 12: 1. you shouldn't use `/app`  but `${FLATPAK_DEST}. 
3. you should also use `-j ${FLATPAK_BUILDER_N_JOBS}` so that the build is done in parallel.

hfiguiere:
Unless you add yourself to the maintainers file and commit to maintain this, it's unlikely to happen.

Erick555:
There is a lot more than libmpv here. I don't know if shared-modules are supposed to have such bundles which may be opinionated. This looks like targeting specific app to me.

plata:
I will bundle it with the app.

PR #233, opened 2023-02-17
Add python3-yt-dlp

plata:


plata:
CC @JBBgameich

A6GibKm:
Could you provide a list of apps that use this module?

plata:
I'm currently only aware of Plasmatube. Maybe @JBBgameich can provide more details on this version compared to the "regular" yt-dlp as used e.g. by io.mpv.Mpv.

hfiguiere:
This is kinda out of scope for shared-modules.

Erick555:
I think it may be too specific for shared-modules. Also I would call it just `yt-dlp` without python3 prefix

Erick555:
I would call this just `yt-dlp` without python3 prefix. The thing that sucks here is the fact update bot doesn't work for shared-modules so for some apps switching to shared-module may be a step backward in maintenance flow (this thing updates very often).

plata:
So just add it to Plasmatube locally?

hfiguiere:
yes.

PR #236, opened 2023-02-26
Add LuaJIT 2.1.0-beta3

ranisalt:
LuaJIT is [used by a lot of packages](https://github.com/search?q=org%3Aflathub%20luajit&type=code), notably Neovim, OBS Studio and a lot of media players (Celluloid, Haruna, MPV, and more).

Includes Anitya checker data. After cleanup, only shared libraries and extensions are left. This is what almost all, if not all, packages are using.

### DIFF
--- a/luajit/luajit.json
+++ b/luajit/luajit.json
@@ -2,6 +2,7 @@
     "name": "luajit",
     "no-autogen": true,
     "make-args": [
+        "BUILDMODE=dynamic",
         "PREFIX=${FLATPAK_DEST}"
     ],
     "make-install-args": [
@@ -23,7 +24,6 @@
     "cleanup": [
         "/bin",
         "/include",
-        "/lib/*.a",
         "/lib/pkgconfig",
         "/share/man"
     ]
--- a/luajit/luajit.json
+++ b/luajit/luajit.json
@@ -1,0 +1,30 @@
+{
+    "name": "luajit",
+    "no-autogen": true,
+    "make-args": [
+        "PREFIX=${FLATPAK_DEST}"
+    ],
+    "make-install-args": [
+        "PREFIX=${FLATPAK_DEST}"
+    ],
+    "sources": [
+        {
+            "type": "archive",
+            "url": "https://luajit.org/download/LuaJIT-2.1.0-beta3.tar.gz",
+            "sha256": "1ad2e34b111c802f9d0cdf019e986909123237a28c746b21295b63c9e785d9c3",
+            "x-checker-data": {
+                "type": "anitya",
+                "project-id": 6444,
+                "stable-only": false,
+                "url-template": "https://luajit.org/download/LuaJIT-$version.tar.gz"
+            }
+        }
+    ],
+    "cleanup": [
+        "/bin",
+        "/include",
+        "/lib/*.a",
+        "/lib/pkgconfig",
+        "/share/man"
+    ]
+}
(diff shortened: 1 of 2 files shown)


TingPing:
This needs a maintainer: Somebody to add themselves to the `CODEOWNERS` file, handle updates, and help update existing packages to use the new shared module.

luajit/luajit.json
```diff
@@ -0,0 +1,30 @@
...
+                "project-id": 6444,
+                "stable-only": false,
+                "url-template": "https://luajit.org/download/LuaJIT-$version.tar.gz"
+            }
+        }
+    ],
+    "cleanup": [
+        "/bin",
+        "/include",
+        "/lib/*.a",
```
TingPing, line 26: Does this actually build a `.a` file? Can you disable that at build time?

ranisalt: It does, but I'll see if that can be disabled. Plain autotools is hard to navigate 😕

ranisalt:
@TingPing I would gladly do it myself, but I'm not a member of the organization, is that OK?

luajit/luajit.json
TingPing: Oh I wished Luajit used autotools. No they hand roll terrible Makefiles.

TingPing: I don't see an option for it.

You can just do:

```json
"post-install": [
    "rm $FLATPAK_DEST/lib/luajit*.a"
]
```

The reason I ask this is because if the `.a` is installed then it is possible for an application to accidentally link to that one.

TingPing:
@ranisalt Yes, you will be added to this repo to be able to review PRs against luajit.

luajit/luajit.json
ranisalt: > I don't see an option for it.

Just found it, if I pass `BUILDMODE=dynamic` to `make` it does not compile the static library.

Erick555:
@ranisalt note that Anitya checker data doesn't seem to do anything in this repo, i.e. it never triggered any update for [gzdoom](https://github.com/flathub/shared-modules/blob/master/gzdoom/gzdoom.json) modules. That means updating this may be pure manual job and actually step backwards for apps.

ranisalt:
> @ranisalt note that Anitya checker data doesn't seem to do anything in this repo, i.e. it never triggered any update for [gzdoom](https://github.com/flathub/shared-modules/blob/master/gzdoom/gzdoom.json) modules. That means updating this may be pure manual job and actually step backwards for apps.

Great, thanks for the tip. I'll try to use the GitHub mirror instead: https://github.com/LuaJIT/LuaJIT

Erick555:
@ranisalt no, I meant **shared-modules** github repo (this one). The flathub bot doesn't work here. You can't fix it yourself.

TingPing:
I'd expect applications to still get the notification?

Indeed this repo never will.

Erick555:
> I'd expect applications to still get the notification?

I don't think so. The update must be done in shared-modules git repo so the bot can't open PR in different repo. This wouldn't make sense.

TingPing:
Hmm, it would be nice to improve this.

Ideally:

- We have a custom Action for this repo that checks versions
- The normal fedc becomes shared-module aware and does submodule updates.

TingPing:
I believe this is all very unimportant for luajit. They don't have releases often.

orowith2os:
Should I look into adding an update checker using GitHub workflows for shared-modules anyways? Seems like it would be useful, particularly for SDL2 and some commonly used modules.

TingPing:
@orowith2os Yeah that would be great.

PR #237, opened 2023-03-05
Add an external-data-checker workflow

orowith2os:
We'll need to figure out how to handle several manifests at once. Should there be a separate workflow for each, so that they can customize how often they run, such as if a module needs to update more often than another? Does the data checker allow several manifests as input?

Cc @TingPing

### DIFF
--- a/.github/workflows/update.yaml
+++ b/.github/workflows/update.yaml
@@ -1,0 +1,30 @@
+name: Check for updates
+on:
+  schedule:
+  - cron: "0 0 * * *" # run every day, since we don't need to run every hour like a normal Flatpak.
+  workflow_dispatch: # can be manually dispatched under GitHub's "Actions" tab 
+
+jobs:
+  flatpak-external-data-checker:
+    runs-on: ubuntu-latest
+
+    strategy:
+      matrix:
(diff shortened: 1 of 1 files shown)


TingPing:
If it is possible I think we should have a single workflow that checks all modules, then opens a separate PR for each module with updates (if a PR was not already opened).

I'm not super familiar with the data checker tool itself, it may need modifications.

orowith2os:
I don't believe having one workflow check all manifests and open one PR for each manifest is possible... But maybe the external-data-checker folks can figure something out?

TingPing:
It would probably involve some custom scripts, but I think an action can use the Github API however it wishes.

orowith2os:
I'll set up a custom workflow based on the data checker at some point. 

It's going to involve setting all the manifests to check inside of a file and running a script that goes through that file. This is probably the cleanest solution. I might also set up a build workflow to be sure everything works correctly.

orowith2os:
Going to pick this up again, but use a more custom workflow. The one we normally use on apps isn't suitable here.

orowith2os:
See https://github.com/flathub/shared-modules/pull/250

PR #238, opened 2023-03-05
Anitya data checker for SDL2

orowith2os:
This only adds a checker for sdl2-no-libdecor, ideally the with-libdecor manifest would reference no-libdecor so that there isn't any duplicated version checking. I'm unable to test it on my end currently, so someone else testing it would be appreciated.

If with-libdecor *can't* reference no-libdecor, I'll add the anitya checker to it.

### DIFF
--- a/SDL2/SDL2-no-libdecor.json
+++ b/SDL2/SDL2-no-libdecor.json
@@ -6,7 +6,13 @@
     {
       "type": "archive",
       "url": "https://github.com/libsdl-org/SDL/archive/refs/tags/release-2.26.3.tar.gz",
-      "sha256": "c661205a553b7d252425f4b751ff13209e5e020b876bbfa1598494af61790057"
+      "sha256": "c661205a553b7d252425f4b751ff13209e5e020b876bbfa1598494af61790057",
+      "x-checker-data": {
+          "type": "anitya",
+          "project-id": 4779,
+          "versions": {"<": "3"},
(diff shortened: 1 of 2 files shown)


TingPing:
I think you'll just have to add the information to both manifests.

orowith2os:
*sigh* I'll get on that then. I'll merge this when I test it on a local machine, or if someone else makes sure it works.

ghisvail:
Also, it looks like the sha256 changed to `af0ff86e4a268bc12c915a0ea19ea1a16419ab1426d92bad5f4e26eb31583967`.

ghisvail:
I see where I made the mistake. The assets include a [tarball](https://github.com/libsdl-org/SDL/releases/download/release-2.26.3/SDL2-2.26.3.tar.gz) too, which has got a different checksum than the one [generated](https://github.com/libsdl-org/SDL/archive/refs/tags/release-2.26.3.tar.gz) by GitHub releases. I wonder whether we should be using that one instead (which is signed too, so it is probably considered official)?

Besides, version 2.26.4 has been released, so we should be able to test whether release monitoring works fairly quickly.

Erick555:
Yes, the upstream made tarballs should be used instead of github made ones. Recently there was a lot of controversy about the latter when github internal changes invalidated all checksums. They reverted the change in the end but this shows the risks.

orowith2os:
See https://github.com/flathub/shared-modules/pull/250

PR #239, opened 2023-03-11
update glu archive URL

christianrauch:
The old URL gives me:
```
Downloading sources
Downloading https://mesa.freedesktop.org/archive/glu/glu-9.0.2.tar.xz
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100   162  100   162    0     0    268      0 --:--:-- --:--:-- --:--:--   268
  0     0    0     0    0     0      0      0 --:--:--  0:00:01 --:--:--     0
Failed to download sources: module glu: server certificate verification failed. CAfile: none CRLfile: none
```
when using `flatpak-builder` locally, and:
```
$ curl https://mesa.freedesktop.org/archive/glu/glu-9.0.3.tar.xz 
<html>
<head><title>301 Moved Permanently</title></head>
<body>
<center><h1>301 Moved Permanently</h1></center>
<hr><center>nginx</center>
</body>
</html>
```

### DIFF
--- a/glu/glu-9.json
+++ b/glu/glu-9.json
@@ -4,7 +4,7 @@
   "sources": [
     {
       "type": "archive",
-      "url": "https://mesa.freedesktop.org/archive/glu/glu-9.0.2.tar.xz",
+      "url": "https://archive.mesa3d.org/glu/glu-9.0.2.tar.xz",
       "sha256": "6e7280ff585c6a1d9dfcdf2fca489251634b3377bfc33c29e4002466a38d02d4"
     }
   ],


ranisalt:
A better option would be `https://archive.mesa3d.org/glu/glu-9.0.2.tar.xz`, the URL that the current URL redirects to, which would not require changing the checksum & archive type.

hfiguiere (changes requested):
As mentionned there is a better alternative.

christianrauch:
Is updating the checksum a problem?

The source is directly from the upstream repo at https://gitlab.freedesktop.org/mesa/glu. This is where I would expect to find the original source.

ranisalt:
The source from mesa3d.org has signed downloads: https://archive.mesa3d.org/glu/ so the checksum can be verified and trusted. The download from freedesktop gitlab does not display a checksum for checking, you have to download it and calculate, which is too late.

christianrauch:
> The source from mesa3d.org has signatures: https://archive.mesa3d.org/glu/ so the checksum can be verified and trusted. The download from freedesktop gitlab does not display a checksum for checking, you have to download it and calculate, which is too late.

Ah, thanks. That makes sense. So I updated to URL only now.

But, how do you extract the checksum from the signature (`glu-9.0.2.tar.xz.sig`)? I guess you can use the signature to verify the file, but I cannot find how to just get the sha256.

ranisalt:
> But, how do you extract the checksum from the signature (`glu-9.0.2.tar.xz.sig`)? I guess you can use the signature to verify the file, but I cannot find how to just get the sha256.

As far as I'm concerned, you can't, but if the signature verifies correct, you can be sure that the file is correct and has not been manipulated. You can then calculate the checksum with `sha256sum` (or other variants) or using `gpg` with `--print-md <algo>` or `--print-mds` which will calculate several different hashes.

ranisalt (approved):

christianrauch:
> > But, how do you extract the checksum from the signature (`glu-9.0.2.tar.xz.sig`)? I guess you can use the signature to verify the file, but I cannot find how to just get the sha256.
> 
> As far as I'm concerned, you can't, but if the signature verifies correct, you can be sure that the file is correct and has not been manipulated. You can then calculate the checksum with `sha256sum` (or other variants) or using `gpg` with `--print-md <algo>` or `--print-mds` which will calculate several different hashes.

But then the argument about not having to change the checksum does not make sense. If you cannot verify the download via the signature file in flatpak builder and have to download and compute the checksum manually anyway, then it does not matter which file you download. I.e. how do you know that the person adding the sha256 did verify the archive via the signature? If the build tool does not verify the signature, then it does not matter if you use the archive with or without a signature.

Issue #234, opened 2023-02-17
Shared module request: libnotify


Mikaela:
While libnotify is part of `org.gnome.Platform`, I am seeing many `org.freedesktop.Platform` using flatpaks include it manually and I think it would be a good candidate for shared-modules.

[I asked about this on Matrix and was encouraged to open an issue here](https://matrix.to/#/%23flatpak%3Amatrix.org/%24ftzwQp8CurbQLgb07P0JWyKtsOCx6hNhm7njdOgapRA?via=pikaviestin.fi&via=matrix.org&via=knot.im&via=envs.net).

A6GibKm:
I think there are about 27 modules using it. I think this is a reasonable proposal for a shared-module but the hard part is having someone do the PR,  maintain  it for the foreseeable future, and open ~27 MRs to modules to projects building it manually.

Regarding the last point, from 0.8 libnotify onwards it uses the notification portal so one should remove the `--talk-name=org.freedesktop.Notifications` permissions.

orowith2os:
Why not add it to the freedesktop runtime instead?

A6GibKm:
And some apps do not even work with the regular install, protonmail-bridge requires either an old version or to disable the gcrypt option to work. Other apps had similar problems.

Erick555:
What gcrypt has to do with libnotify?

A6GibKm:
I am sorry, I was just in a convo about libsecret and I mixed the two of them in my mind.

Regarding the discussion, I have seen very little use of libnotify outside of GNOME to be honest, and it GNOME it makes no sense to use it anyways.

bbhtt:
https://gitlab.com/freedesktop-sdk/freedesktop-sdk/-/issues/1642 If it doesn't get included in runtime I'll add it to shared-modules and handles reaching out to the ~30-33 apps using it.

nanonyme:
I would note that 23.08 is already out and the request to include it into runtime is already months late.

bbhtt:
It was added to the SDK/runtime. If everything goes well, it'll be in 24.08 I think.

PR #243, opened 2023-04-03
Add icoutils 0.32.3

ranisalt:
#### Summary

`icoutils` is a set of command-line tools used for working with Windows icon and cursor files. It can create, extract, and convert icons and cursors in different formats, and is useful for developers working with these resources in their applications.

It is currently used by 5 flathub packages. `icoutils` could be used in others to extract icons from Windows executables instead of providing their own icons.

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -23,3 +23,4 @@
 /gzdoom/           @Eonfge
 /vorbisgain/       @Eonfge
 /luajit/           @ranisalt
+/icoutils/         @ranisalt
--- a/icoutils/icoutils.json
+++ b/icoutils/icoutils.json
@@ -1,0 +1,15 @@
+{
+  "name": "icoutils",
+  "sources": [
+    {
+      "type": "archive",
+      "url": "https://savannah.nongnu.org/download/icoutils/icoutils-0.32.3.tar.bz2",
+      "sha256": "17abe02d043a253b68b47e3af69c9fc755b895db68fdc8811786125df564c6e0",
+      "x-checker-data": {
+        "type": "anitya",
+        "project-id": 1360,
+        "url-template": "https://savannah.nongnu.org/download/icoutils/icoutils-$version.tar.bz2"
+      }
+    }
+  ]
+}


orowith2os:
Can you give me a list of all the packages that use this? I'd love to see how they use it, but this seems pretty niche in the first place; if the packages using it are Wine-based, it could make sense to include it in the baseapp instead.

ranisalt:
@orowith2os sure! Note that "Windows icon" is the `.ico` file, not necessarily a Windows-only feature.

- [ghidra](https://github.com/flathub/org.ghidra_sre.Ghidra), [openmsx](https://github.com/flathub/org.openmsx.openMSX), [teeworlds](https://github.com/flathub/com.teeworlds.Teeworlds) use it to extract an icon from a .ico file
- [dolphin](https://github.com/flathub/org.kde.dolphin) uses it during runtime to get icon thumbnails
- [q4wine](https://github.com/flathub/ua.org.brezblock.q4wine) uses it but it's kinda wine based

Erick555:
So except q4wine those aren't wine apps which means adding this to wine bease app wouldn't help.

On other hand this app list is quite short and some of them aren't even actively maintained. Also others use f-e-d-c to auto update icoutils and moving it to shared-modules will only require more effort rather than less.

I think there is little to gain from having this in shared-modules.

A6GibKm:
From q4wine, the manifest has
```
      - name: icoutils
        sources:
          - type: archive
            url: http://savannah.nongnu.org/download/icoutils/icoutils-0.32.3.tar.bz2
            sha256: 17abe02d043a253b68b47e3af69c9fc755b895db68fdc8811786125df564c6e0
```

this seems rather easy to add to projects

ranisalt:
Right, but what is the cut for a module to make into this repo?

mac and libsoup are also quite simple and used by even less packages (4 and 3, respectively) than icoutils. Notably some recipes are not using the shared module for libsoup so the count should be higher. Impossible to search for "mac" 😂

orowith2os:
I've updated quite a few apps to the GNOME 44 runtime and many of them, if they do anything network-wise, have complained about libsoup. There are more users of that than you think.

Not sure about mac.

If it's used by many apps, I'd say 10-15 range, it could be added as a submodule imo. But that depends on the module and their use cases. That's why it helps to discuss whether it should be added.

ranisalt:
Sounds good to me!

Erick555:
> If it's used by many apps, I'd say 10-15 range

27 apps was considered [not much](If it's used by many apps, I'd say 10-15 range) recently. Just sayin

ranisalt:
> 27 apps was considered [not much](https://github.com/flathub/shared-modules/issues/234#issuecomment-1472846819) recently. Just sayin

I think it's a different scenario. 27 apps is not too much to add to _every single app out there_, but might be enough to add to this optional, no disk space occupied if not used, module repo, for the sake of avoiding repetition

TingPing:
It is hard to set a simple number to cross. But I think it is safe to say 5 isn't it.

In my mind the value of libsoup is that it *was* in the GNOME runtime and removed. It being here gives a clear upgrade path for some applications.

PR #245, opened 2023-04-22
intltool: Add checker data

dreua:


TingPing:
intltool is a dead project. The last release was in 2015 (and 2012 before that).

dreua:
Does that mean it should not be used any more? Do you happen to know what it should be replaced with?

ranisalt:
> Do you happen to know what it should be replaced with?

I have seen it being replaced by `gettext`

hfiguiere:
that just mean that it won't be updated.

if a project needs it, its here.

dreua:
Software that won't be updated shouldn't be depended upon imo. If it is used (and widely enough that it justifies being a module here) then there might be circumstances where there could be an update, even if the project hasn't been actively maintained for some years. An in this case I would like my external data checker to notify me, that's why I added the checker data. I don't see the harm in adding it *just in case* there will be another release.

If we are 100% sure that there won't be updates then everyone should be migrating away from this dead project.

TingPing:
Upstream software should use regular `gettext`. Please do help projects move to it.

This shared-module exists because it is commonly used by older projects.

`intltool` is a relatively simple set of perl scripts that are used at build time. So the "risks" involved are low (no security problems, perl will never change, etc).

It is just a waste of time to check updates for it. It is simply dead.

Issue #201, opened 2022-08-07
pygame fails to compile


bbhtt:
With GNOME 42 SDK:

```
error
  error: subprocess-exited-with-error
  
  × python setup.py bdist_wheel did not run successfully.
  │ exit code: 1
  ╰─> [501 lines of output]

...

      src_c/imageext.c: In function ‘png_write_fn’:
      src_c/imageext.c:217:45: warning: comparison of integer expressions of different signedness: ‘int’ and ‘png_size_t’ {aka ‘long unsigned int’} [-Wsign-compare]
        217 |     if (SDL_RWwrite(rwops, data, 1, length) != length) {
            |                                             ^~
      src_c/imageext.c: In function ‘j_term_destination’:
      src_c/imageext.c:488:68: warning: comparison of integer expressions of different signedness: ‘int’ and ‘size_t’ {aka ‘long unsigned int’} [-Wsign-compare]
        488 |         if (SDL_RWwrite(dest->outfile, dest->buffer, 1, datacount) != datacount) {
            |                                                                    ^~
      src_c/imageext.c: In function ‘write_jpeg’:
      src_c/imageext.c:559:32: warning: comparison of integer expressions of different signedness: ‘int’ and ‘JDIMENSION’ {aka ‘unsigned int’} [-Wsign-compare]
        559 |         if (num_lines_to_write >
            |                                ^
      In file included from src_c/pygame.h:32,
                       from src_c/imageext.c:29:
      src_c/imageext.c: In function ‘opengltosdl’:
      src_c/_pygame.h:331:47: warning: right-hand operand of comma expression has no effect [-Wunused-value]
        331 | #define RAISE(x, y) (PyErr_SetString((x), (y)), (PyObject *)NULL)
            |                     ~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~
      src_c/imageext.c:718:9: note: in expansion of macro ‘RAISE’
        718 |         RAISE(PyExc_RuntimeError, "Cannot get video surface.");
            |         ^~~~~
      src_c/_pygame.h:331:47: warning: right-hand operand of comma expression has no effect [-Wunused-value]
        331 | #define RAISE(x, y) (PyErr_SetString((x), (y)), (PyObject *)NULL)
            |                     ~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~
[...]

PR #247, opened 2023-06-04
Add pygame 2

sparshg:
Since pygame 2 is based on SDL2, its dependencies are also added.


A6GibKm:
Could you please give some examples of apps that make use of this module as bundled here?

sparshg:
Currently we are updating the flatpak packages like [Physics](https://flathub.org/apps/org.sugarlabs.Physics) and [Pippy](https://flathub.org/apps/org.sugarlabs.Pippy) which require this module. We are also planning to add more packages which may use pygame.

SDL2/SDL2_mixer-2.6.3.json
```diff
@@ -0,0 +1,22 @@
+{
+    "name": "SDL2_mixer",
```
hfiguiere, line 2: this is in the freedesktop SDK. You don't need it.

SDL2/SDL2_ttf-2.20.2.json
```diff
@@ -0,0 +1,22 @@
+{
+    "name": "SDL2_ttf",
```
hfiguiere, line 2: 2.20.0 is in the Freedesktop SDK

SDL2/SDL2_image-2.6.3.json
```diff
@@ -0,0 +1,19 @@
+{
+    "name": "SDL2_image",
```
hfiguiere, line 2: This is in the Freedesktop SDK

hfiguiere (changes requested):
Don't add unecessary modules

pygame2/pygame-2.4.0.json
```diff
@@ -0,0 +1,42 @@
...
+            "arm": {
+                "env": {
+                    "ORIGLIBDIRS": "/lib:/lib/arm-linux-gnueabihf"
+                }
+            },
+            "i386": {
+                "env": {
+                    "ORIGLIBDIRS": "/lib:/lib/i386-linux-gnu"
+                }
+            },
```
hfiguiere, line 34: neither of these arch is supported.

Erick555:
The person who adds shared-module is expected to maintain it for the foreseeable future. They should know the build details and what dependency is necessary.

chimosky:
> The person who adds shared-module is expected to maintain it for the foreseeable future. They should know the build details and what dependency is necessary.

Agreed.

orowith2os (changes requested):
I'm going to have to go with denying this, tbh. It feels better to have in a BaseApp than a submodule, imo.

barthalion:
If it takes less than 30 minutes to build, maintenance overhead of a BaseApp is not worth it.

orowith2os:
> If it takes less than 30 minutes to build, maintenance overhead of a BaseApp is not worth it.

I was thinking a more generic BaseApp, that just includes some common game libraries (pygame, newer SDL2, wine, gamemode). This would also solve https://github.com/flathub/shared-modules/pull/221.

A6GibKm:
First you need to define a clear usecase, are these apps using wine for example? We already have a base app with wine.

pygame2/pygame-2.4.0.json
```diff
@@ -0,0 +1,32 @@
...
+    "sources": [
+        {
+            "type": "archive",
+            "url": "https://files.pythonhosted.org/packages/b9/42/154038c4988d750ded799dc5d6ca4e631db93888e3d76084625b0588dfd2/pygame-2.4.0.tar.gz",
+            "sha256": "e3603e70e96ee30af1954ce57d4922a059402f368013e7138e90f1c03d185267"
+        }
+    ],
+    "buildsystem": "simple",
+    "build-commands": [
+        "pip3 install --ignore-installed --no-deps --prefix=/app ."
```
bbhtt, line 12: You should run `setup.py cython`, to recythonise the sources before installing. Upstream c sources might not match with the python api available in the runtime or app. The c api time to time introduces changes leading to build failures like https://github.com/flathub/shared-modules/issues/201 and the python version is a moving target.

https://www.pygame.org/wiki/GettingStarted#Compilation
> Some of the .c files are generated by [Cython](https://cython.org/) from .pyx files. Running "setup.py cython" will update them.

hfiguiere:
You are missing the addition to `CODEOWNERS` to commit to maintain it.

PR #250, opened 2023-06-21
Add data checker workflow and some data checkers

orowith2os:
Supersedes https://github.com/flathub/shared-modules/pull/237 and https://github.com/flathub/shared-modules/pull/238

Allows closing https://github.com/flathub/shared-modules/issues/92

Would like a review from (2-3) others before merging.

### DIFF
--- a/.github/workflows/update.yaml
+++ b/.github/workflows/update.yaml
@@ -1,7 +1,7 @@
 name: Check for updates
 on:
   schedule: # for scheduling to work this file must be in the default branch
-    - cron: "0 * * * *" # run every hour
+    - cron: "0 0 * * 0" # 00:00 on every Sunday
   workflow_dispatch: # can be manually dispatched under GitHub's "Actions" tab
 
 jobs:
@@ -13,7 +13,7 @@ jobs:
         branch: [ master ] # list all branches to check
 
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
         with:
           ref: ${{ matrix.branch }}
 
--- a/.github/workflows/update.yaml
+++ b/.github/workflows/update.yaml
@@ -1,0 +1,31 @@
+name: Check for updates
+on:
+  schedule: # for scheduling to work this file must be in the default branch
+    - cron: "0 * * * *" # run every hour
+  workflow_dispatch: # can be manually dispatched under GitHub's "Actions" tab
+
+jobs:
+  flatpak-external-data-checker:
+    runs-on: ubuntu-latest
+
+    strategy:
+      matrix:
+        branch: [ master ] # list all branches to check
+
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          ref: ${{ matrix.branch }}
+
+      - uses: docker://ghcr.io/flathub/flatpak-external-data-checker:latest
+        env:
+          GIT_AUTHOR_NAME: Flatpak External Data Checker
+          GIT_COMMITTER_NAME: Flatpak External Data Checker
+          # email sets "github-actions[bot]" as commit author, see https://github.community/t/github-actions-bot-email-address/17204/6
+          GIT_AUTHOR_EMAIL: 41898282+github-actions[bot]@users.noreply.github.com
+          GIT_COMMITTER_EMAIL: 41898282+github-actions[bot]@users.noreply.github.com
+          EMAIL: 41898282+github-actions[bot]@users.noreply.github.com
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        with:
+          # We don't want the default entrypoint, we want our own script
+          entrypoint: update-checker.sh
--- a/README.md
+++ b/README.md
@@ -54,7 +54,7 @@ We provide an automatic updating mechanism for submodules located here. In order
 
 - Set up x-checker-data for your sources: https://github.com/flathub/flatpak-external-data-checker#url-checker
 
-And it will check for an update every hour.
+And it will check for updates in a weekly basis.
 
 Please do not request adding modules unless they are widely used in
 the Flathub repository.
--- a/README.md
+++ b/README.md
@@ -50,6 +50,11 @@ git rm -f shared-modules
 rm .gitmodules
 ```
 
+We provide an automatic updating mechanism for submodules located here. In order to utilize it:
+
+- Set up x-checker-data for your sources: https://github.com/flathub/flatpak-external-data-checker#url-checker
+
+And it will check for an update every hour.
 
 Please do not request adding modules unless they are widely used in
 the Flathub repository.
--- a/update-checker.sh
+++ b/update-checker.sh
@@ -1,0 +1,36 @@
+#!/bin/bash
+
+# Args to pass to the data checker
+args=("--update" "--never-fork")
+
+# Initialize an array to store file paths
+file_paths=()
+
+# Iterate over directories next to the script
+for dir in ./*/; do
+  # Search for files with the extensions .yaml, .yml, and .json
+  files=("$dir"*.yaml "$dir"*.yml "$dir"*.json)
+
+  # Iterate over the files
+  for file in "${files[@]}"; do
+    # Check if the file exists and contains "x-checker-data"
+    if [[ -f "$file" ]] && grep -q "x-checker-data" "$file"; then
+      # Add the file path to the array
+      file_paths+=("$file")
+    fi
+  done
+done
+
+# Print the file paths in the array
+for path in "${file_paths[@]}"; do
+  echo "Running data checker on: $path"
+
+ # If we're not running in a container, use the Flatpak. Else, assume we're
+ # running in a container and call the data checker directly.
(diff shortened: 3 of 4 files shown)


update-checker.txt
```diff
@@ -0,0 +1,4 @@
+# List of all the files to check with the data checker
```
barthalion, line 1: There's probably no need to make this opt-in. Just grep for x-data-checker across the repo to get the list of modules that "signed up".

wjt (commented):
Very nice! I Agree With Bart.

.github/workflows/update.yaml
```diff
@@ -0,0 +1,32 @@
+name: Check for updates
+on:
+  schedule: # for scheduling to work this file must be in the default branch
+  - cron: "0 * * * *" # run every hour
```
ranisalt, line 4:
```suggestion
    - cron: "0 * * * *" # run every hour
```

To keep the formatting consistent

update-checker.sh
```diff
@@ -0,0 +1,38 @@
...
+  fi
+
+  echo "Running data checker on: $line"
+
+ # If we're not running in a container, use the Flatpak. Else, assume we're
+ # running in a container and call the data checker directly.
+  if [[ ! -f /run/.containerenv && ! -f /.dockerenv ]]; then
+   flatpak run --filesystem="$(pwd)" org.flathub.flatpak-external-data-checker $line
+  else
+   /app/flatpak-external-data-checker --update --never-fork $line
```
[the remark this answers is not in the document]
orowith2os: The pwd? It should be, from what I've seen with other workflows and whatnot. 

As Bart mentioned, I should probably just grep for the data checker instead, so I can get the *real* paths and not trust this file.

orowith2os:
Oh dear, the commit history is a mess. I'll make sure to squash these when I get some approvals.

orowith2os:
I've tested with gzdoom and the SDL2 manifests, all appears to be working well.

ranisalt:
> Oh dear, the commit history is a mess. I'll make sure to squash these when I get some approvals.

Don't worry, it can be done by the merger. Great job!

orowith2os:
@barthalion @hfiguiere @wjt lil reminder, would you mind giving me your thoughts on this? Am I good to merge?

README.md
```diff
@@ -42,6 +42,13 @@ git rm -f shared-modules
 rm .gitmodules
 ```
 
+We provide an automatic updating mechanism for submodules located here. In order to utilize it:
+
+1. Set up x-checker-data for your sources: https://github.com/flathub/flatpak-external-data-checker#url-checker
+2. Add the path to your sources to `update-checker.txt`
+
+And it will update automatically.
```
barthalion, line 50: This is out of date now, isn't it?

.github/workflows/update.yaml
```diff
@@ -0,0 +1,32 @@
...
+          GIT_COMMITTER_NAME: Flatpak External Data Checker
+          # email sets "github-actions[bot]" as commit author, see https://github.community/t/github-actions-bot-email-address/17204/6
+          GIT_AUTHOR_EMAIL: 41898282+github-actions[bot]@users.noreply.github.com
+          GIT_COMMITTER_EMAIL: 41898282+github-actions[bot]@users.noreply.github.com
+          EMAIL: 41898282+github-actions[bot]@users.noreply.github.com
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        with:
+          # We don't want the default entrypoint, we want our own script :p
+          entrypoint: update-checker.sh
+          args: update-checker.txt
```
wjt, line 32: I think this argument is now unneeded?

.github/workflows/update.yaml
```diff
@@ -0,0 +1,32 @@
...
+          GIT_AUTHOR_NAME: Flatpak External Data Checker
+          GIT_COMMITTER_NAME: Flatpak External Data Checker
+          # email sets "github-actions[bot]" as commit author, see https://github.community/t/github-actions-bot-email-address/17204/6
+          GIT_AUTHOR_EMAIL: 41898282+github-actions[bot]@users.noreply.github.com
+          GIT_COMMITTER_EMAIL: 41898282+github-actions[bot]@users.noreply.github.com
+          EMAIL: 41898282+github-actions[bot]@users.noreply.github.com
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        with:
+          # We don't want the default entrypoint, we want our own script :p
+          entrypoint: update-checker.sh
```
wjt, line 31: Does this actually work? Do you not need:

```suggestion
          entrypoint: ./update-checker.sh
```

(I have no idea either way.)

wjt (commented):
Seems OK, except for a nitpick similar to Bart's

update-checker.sh
```diff
@@ -0,0 +1,36 @@
...
+    if [[ -f "$file" ]] && grep -q "x-checker-data" "$file"; then
+      # Add the file path to the array
+      file_paths+=("$file")
+    fi
+  done
+done
+
+# Print the file paths in the array
+for path in "${file_paths[@]}"; do
+  echo "Running data checker on: $line"
```
wjt, line 26:
```
In update-checker.sh line 26:
  echo "Running data checker on: $line"
                                 ^---^ SC2154: line is referenced but not assigned.
```

update-checker.sh
```diff
@@ -0,0 +1,36 @@
...
+
+ # If we're not running in a container, use the Flatpak. Else, assume we're
+ # running in a container and call the data checker directly.
+  if [[ ! -f /run/.containerenv && ! -f /.dockerenv ]]; then
+   flatpak run --filesystem="$(pwd)" org.flathub.flatpak-external-data-checker ${args[@]} $path
```
wjt, line 31:
```
In update-checker.sh line 31:
   flatpak run --filesystem="$(pwd)" org.flathub.flatpak-external-data-checker ${args[@]} $path
                                                                               ^--------^ SC2068: Double quote array expansions to avoid re-splitting elements.
                                                                                          ^---^ SC2086: Double quote to prevent globbing and word splitting.

Did you mean: 
   flatpak run --filesystem="$(pwd)" org.flathub.flatpak-external-data-checker ${args[@]} "$path"
```

wjt: I think what you want is:

```suggestion
   flatpak run --filesystem="$(pwd)" org.flathub.flatpak-external-data-checker "${args[@]}" "$path"
```

update-checker.sh
```diff
@@ -0,0 +1,36 @@
...
+  else
+   /app/flatpak-external-data-checker ${args[@]} $path
```
wjt, line 33:
```
In update-checker.sh line 33:
   /app/flatpak-external-data-checker ${args[@]} $path
                                      ^--------^ SC2068: Double quote array expansions to avoid re-splitting elements.
                                                 ^---^ SC2086: Double quote to prevent globbing and word splitting.

Did you mean: 
   /app/flatpak-external-data-checker ${args[@]} "$path"
```

```suggestion
   /app/flatpak-external-data-checker "${args[@]}" "$path"
```

update-checker.sh
wjt:
```suggestion
  echo "Running data checker on: $path"
```

wjt (changes requested):
I ran Shellcheck over the script.

For more information:
  https://www.shellcheck.net/wiki/SC2068 -- Double quote array expansions to ...
  https://www.shellcheck.net/wiki/SC2154 -- line is referenced but not assign...
  https://www.shellcheck.net/wiki/SC2086 -- Double quote to prevent globbing ...

.github/workflows/update.yaml
orowith2os: I think what it already has is correct, going off of the GitHub docs?

orowith2os:
Shellcheck says it's okay, I've updated the README, and the workflow appears to be correct, going off of the [GitHub docs](https://docs.github.com/en/actions/creating-actions/metadata-syntax-for-github-actions#runsentrypoint), I think?

.github/workflows/update.yaml
wjt: OK!

wjt (approved):
I don't have commit access, but this LGTM

orowith2os:
I think this is going to need more thought and work, and probably some porting to Python - I want to account for cases where submodules rely on generated dependencies too, and it's also possible that the layout of some submodules isn't what this checker expects. 

This could also do with some work in the main external-data-checker too.

bbhtt:
I would merge this as is now. I don't mind having it in shell and future improvements or python ports can be done later.

Two things - 

1.  Commits should be squashed
2. The workflow run should be weekly, x-checker doesn't have super active maintenance and I don't think piling PRs is a good idea.

bbhtt:
Handled both, let's merge this.

bbhtt:
Seems to be working after a minor fix: https://github.com/flathub/shared-modules/commit/8d84ae268a653163f689e41d9ab441a671754510

https://github.com/flathub/shared-modules/actions/runs/11093647389

These were the first two PRs:

https://github.com/flathub/shared-modules/pull/321

https://github.com/flathub/shared-modules/pull/320

PR #255, opened 2023-08-21
Cleanup /lib/debug

slembcke:
These debug files were a few dozen MB in my build. Cleanup time!

### DIFF
--- a/SDL2/SDL2-with-libdecor.json
+++ b/SDL2/SDL2-with-libdecor.json
@@ -11,6 +11,7 @@
   ],
   "cleanup": [ "/bin/sdl2-config", 
                "/include", 
+               "/lib/debug",
                "/lib/libSDL2.la", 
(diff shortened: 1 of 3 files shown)


Erick555:
Flathub bans cleaning [/lib/debug](https://github.com/flathub/flatpak-builder-lint/blob/1e38c640d506c98fee283d5393d85130a1216ef2/flatpak_builder_lint/checks/toplevel.py#L28). flatpak-builder is supposed to move `/lib/debug` content into .Debug extension automatically. Check why it didn't work for you.

hfiguiere:
No.

The debug symbols are separated to the .Debug package. It's an error to remove them.

slembcke:
Ah, got it. I didn't realize that was done at packaging time. There was an instruction to make sure and clean up any unused files, and I figured maybe a build option got changed when the libraries were updated or something. Thanks!

PR #260, opened 2023-09-09
Update SDL2 version.

SpartanJ:
I need to update the SDL2 version provided by flathub given that the current version has some bugs in Wayland sessions that could potentially break many applications.
[Here's](https://github.com/SpartanJ/ecode/issues/102) the current issue that I cannot avoid with the current version and I verified that it's fixed in the latest version.


orowith2os:
The applications relying on this submodule should also now use the 23.08 runtime, and stop building this; the necessary bits, including libdecor, are in 23.08 now.

PR #262, opened 2023-09-10
Add libnotify

bbhtt:
Closes https://github.com/flathub/shared-modules/issues/234

Draft because waiting on https://gitlab.com/freedesktop-sdk/freedesktop-sdk/-/issues/1642

I kept the x-checker because it makes easier to update and in the future shared-modules might get a x-checker workflow.

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -10,6 +10,7 @@
 /libcanberra/      @hadess
 /libdecor/         @orowith2os
 /SDL2/             @orowith2os
+/libnotify/        @bbhtt
 /libsecret/        @Lctrs
 /libusb/           @A6GibKm
 /openjpeg/         @mbridon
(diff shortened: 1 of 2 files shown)


hfiguiere:
you should add yourself to the maintainers for that module. That's the commitment when you submit it.

hfiguiere:
See `CODEOWNERS`

bbhtt:
Ah thanks forgot about that

orowith2os:
If the runtime gets libnotify (even if not in 23.08) it would probably make sense to close this in favor of waiting until 24.08. So keep that in mind - 30-40 apps shipping this manually isn't a big deal if it means they can strip it later.

Erick555:
If libnotify would be in fsdk 24.08 then it may be not worth adding it here for just one year.

bbhtt:
Yea this is incase if it doesn't get accepted into the runtime. I'll close if it gets in.

bbhtt:
It was added to the SDK/runtime. If everything goes well, it'll be in 24.08 I think. So closing.

PR #266, opened 2023-09-25
Update libdecor to 0.2.0

ColinKinloch:
Is replacing the 0.1.1 file is correct, rather than just adding 0.2.0 alongside?


orowith2os:
Yep, it's correct.

Erick555:
I think it should be named `libdecor.json` without version otherwise each time app updates submodules you need to adjust path manually which is annoying and strong reason to avoid updates.

orowith2os:
We're going to remove this when 22.08 goes end of life anyways, and I had some reason that I forgot for versioning it.

PR #271, opened 2023-09-27
Update sdl-compat to 1.2.68

Eonfge:


Eonfge:
@orowith2os Are you also in charge of SDL/ or only SDL2/ ?

PR #272, opened 2023-09-29
add SDL2_ttf

akallabeth:


Erick555:
This is in runtime. Why do you need it?

akallabeth:
oh, indeed.
did use it previously but missed it is part of runtime now.

PR #275, opened 2023-10-14
Linux audio cleanups

bbhtt:
See commit messages.

### DIFF
--- a/linux-audio/fftw3f.json
+++ b/linux-audio/fftw3f.json
@@ -35,6 +35,7 @@
         "/lib/cmake",
         "/lib/pkgconfig",
         "/share/man",
+        "/share/info",
         "*.la",
         "*.so"
     ]
--- a/linux-audio/lash.json
+++ b/linux-audio/lash.json
@@ -6,6 +6,11 @@
         "--disable-serv-inst",
         "CFLAGS=-D_GNU_SOURCE"
     ],
+    "cleanup": [
+        "/include",
+        "/lib/pkgconfig",
+        "*.la"
+    ],
     "sources": [
         {
             "type": "archive",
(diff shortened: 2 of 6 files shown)


linux-audio/fftw3f.json
```diff
@@ -35,7 +35,7 @@
         "/lib/cmake",
         "/lib/pkgconfig",
         "/share/man",
-        "*.la",
-        "*.so"
```
hfiguiere, line 39: we remove the .so because these are only use for `-l` by the linker. It's always the somajor it points to that's used.

linux-audio/lash.json
```diff
@@ -6,6 +6,11 @@
         "--disable-serv-inst",
         "CFLAGS=-D_GNU_SOURCE"
     ],
+    "cleanup": [
+        "/include",
+        "/lib/pkgconfig",
+        "/lib/*.la"
```
hfiguiere, line 12:
```suggestion
        "*.la"
```

linux-audio/liblo-static.json
```diff
@@ -5,7 +5,6 @@
     "/include",
     "/lib/pkgconfig",
     "*.la",
-    "*.a",
```
hfiguiere, line 8: we remove this too for a good reason. it's a static lib.

linux-audio/lilv.json
```diff
@@ -39,9 +39,11 @@
         "/bin",
         "/include",
         "/lib/pkgconfig",
+        "/share/doc",
         "/share/man"
       ]
     },
+    "lv2.json",
```
hfiguiere, line 46:
```suggestion
```

This is deliberate.

linux-audio/fftw3f.json
Erick555: freedesktop-sdk also removes all .so from Platform.

bbhtt: > these are only use for -l by the linker

Ah yea that makes sense, for a moment I thought cleanup would remove it before finishing hence something linking against it won't find.

PR #276, opened 2023-10-23
libusb: pass `--disable-udev`, fix hotplug

v1993:
I'm using libusb (via libgusb) in my program and this is required to get device detection to work. Niepce also [uses it](https://gitlab.gnome.org/GNOME/niepce/-/blob/master/flatpak/net.figuiere.Niepce.json?ref_type=heads#L42) in their flatpak version.

### DIFF
--- a/libusb/libusb.json
+++ b/libusb/libusb.json
@@ -11,6 +11,10 @@
             "type": "archive",
             "url": "https://github.com/libusb/libusb/releases/download/v1.0.26/libusb-1.0.26.tar.bz2",
             "sha256": "12ce7a61fc9854d1d2a1ffe095f7b5fac19ddba095c259e6067a46500381b5a5"
+        },
+        {
+          "type": "patch",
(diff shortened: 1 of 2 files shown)


v1993:
Regarding the patch - it's an original contribution; you have to enable `--shared=network` to let netlink monitor do its job as hotplug.

hfiguiere:
> Niepce also [uses it](https://gitlab.gnome.org/GNOME/niepce/-/blob/master/flatpak/net.figuiere.Niepce.json?ref_type=heads#L42) in their flatpak version.

Pardon?

This manifest isn't tailored to flathub. Don't use it as an example to follow for flathub. It's not even like the app has had a release.

v1993:
Sorry, sorry, I only meant that as an example of this option being used in the wild, not as a primary justification for its inclusion in this repository.

That said... this is quite likely a misunderstanding on my side nonetheless. If using (system, I suppose?) libudev is completely fine then this PR is redundant.

PR #279, opened 2023-11-04
Update gudev to 238

wjt:
Required to update upower as requested in https://github.com/flathub/org.gnome.PowerStats/issues/11


Erick555:
thx

PR #280, opened 2023-11-09
README: Fix redirected Wiki link

hadess:


TingPing:
Maybe this should just be removed. It doesn't say anything about shared-modules anymore, and what it did say wasn't actually helpful:

> A small number of dependencies that are frequently used by applications have been collected in the [shared-modules repository](https://github.com/flathub/shared-modules), for use by applications being distributed with flathub.

README.md
```diff
@@ -43,6 +43,7 @@ rm .gitmodules
 ```
 
 
-[See the description in the Flathub wiki](https://github.com/flathub/flathub/wiki/App-Requirements#shared-modules) for more information.
-
 Please do not request adding modules unless they have many users in the Flathub repository.
+
+All shared modules manifests in this repository are, and need to be, in the JSON format,
+which is supported by both Flatpak manifest formats, JSON and YAML.
```
orowith2os, line 49: Since when was that decided? My understanding so far is that the only reason we won't use YAML is for consistency. If it seems reasonable, a shared module can use YAML for its manifests.

orowith2os: That was almost six years ago now. The information I have and am basing this off of is a bit more recent, at least a year or so.

orowith2os: per @TingPing, in https://github.com/flathub/shared-modules/pull/218#issuecomment-1343078326:

> I would like to keep it as JSON for consistency. Though both file types work the same way.

If it were consistently used (with reason), I don't see why it shouldn't be allowed.

orowith2os: Sounds like A Reason for me to submit a mass PR to convert all the submodules to YAML, then.

TingPing: Or you could just do nothing because JSON is totally functional and doesn't need changing.

PR #285, opened 2024-01-10
Use openresty/luajit2 to get significantly newer version of luajit.

ProjectSynchro:
As the head developer of Luajit does not believe in the concept of releases, the only way to get a newish version of luajit is to use a snapshot of the repository's HEAD. 

v2.1.ROLLING is the 'newest' tag from the 2.1 series of luajit (dated august 2023), and includes a plethora of fixes and improvements.

### DIFF
--- a/luajit/luajit.json
+++ b/luajit/luajit.json
@@ -10,15 +10,10 @@
     ],
     "sources": [
         {
-            "type": "archive",
-            "url": "https://luajit.org/download/LuaJIT-2.1.0-beta3.tar.gz",
-            "sha256": "1ad2e34b111c802f9d0cdf019e986909123237a28c746b21295b63c9e785d9c3",
-            "x-checker-data": {
-                "type": "anitya",
-                "project-id": 6444,
-                "stable-only": false,
-                "url-template": "https://luajit.org/download/LuaJIT-$version.tar.gz"
-            }
+            "type": "git",
+            "url": "https://luajit.org/git/luajit.git",
+            "tag": "v2.1.ROLLING",
+            "commit": "2090842410e0ba6f81fad310a77bf5432488249a"
         }
     ],
     "cleanup": [
--- a/luajit/luajit.json
+++ b/luajit/luajit.json
@@ -13,7 +13,7 @@
             "type": "git",
             "url": "https://github.com/openresty/luajit2.git",
             "tag": "v2.1-20231117",
-            "commit": "4182d6bf37e9f8d1cb5d6e83b1db66de84b95101"
+            "commit": "4182d6bf37e9f8d1cb5d6e83b1db66de84b95101",
             "x-checker-data": {
                 "type": "git",
                 "tag-pattern": "^v([\d.-]+)$",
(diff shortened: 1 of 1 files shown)


ProjectSynchro:
Source of my above comment: https://luajit.org/download.html

TingPing:
The name is a bit odd, `v2.1.ROLLING` is a tag that will never change?

ranisalt:
> As the head developer of Luajit does not believe in the concept of releases

This is an absolutely stupid stance. _Lunatic_, if you pardon the pun.

> The name is a bit odd, `v2.1.ROLLING` is a tag that will never change?

This is just the branch 2.1 that keeps getting new commits as the maintainer wants. In practice, he abandoned versioning and just expects everyone to use master.

hfiguiere:
so tag + commit will break the build in case of retag.

TingPing:
We can specify a specific commit and bump it ocassionally. It is all around just annoying though.

ProjectSynchro:
I mainly decided to pick that tag since it's at least a newer "release" of luajit.. But yes, the lack of versioning is and has been problematic for a bunch of different distributions, since they don't typically like using git snapshots. Also the concept of releases being "antiquated" is just.. yeah lets not get too into that.

Now: I have recently seen that some of the BSDs, Alpine Linux (and maybe more) as well as macports and a few others have taken to packaging a branch/fork of luajit developed by openresty: https://github.com/openresty/luajit2.

This branch/fork has the benefit of having tags that are "released" (ha) every month or so and could potentially be used to allow flatpak-external-data-checker to notify on an update. See: https://github.com/openresty/luajit2/tags

The openresty project does pull from the 2.1 branch of luajit semi-frequently and appears to include some other goodies, but in my testing it does not break any existing projects that link to luajit 2.1, as it is still luajit "2.1".

Unfortunately with MoonJIT being dead in the water now, it seems like using this openresty fork/branch is the best path forward. Do you all think that using that project as the source for luajit is reasonable?

ProjectSynchro:
I changed the source to [openresty/luajit2](https://github.com/openresty/luajit2), and additionally added support for flatpak-external-data-checker that checks the git tags for updates.

ranisalt:
Works for me if it's always API-compatible with LuaJIT, which seems to be the case.

luajit/luajit.json
```diff
@@ -10,14 +10,14 @@
...
+            "url": "https://github.com/openresty/luajit2.git",
+            "tag": "v2.1-20231117",
+            "commit": "4182d6bf37e9f8d1cb5d6e83b1db66de84b95101",
             "x-checker-data": {
-                "type": "anitya",
-                "project-id": 6444,
-                "stable-only": false,
-                "url-template": "https://luajit.org/download/LuaJIT-$version.tar.gz"
+                "type": "git",
+                "tag-pattern": "^v([\d.-]+)$",
```
TingPing, line 19: Did you test this with the data checker? I assume you need to escape `\\`

ProjectSynchro: Whoops! Sorry! 

Good catch, I had eyeballed the regex but I guess my regex-fu is not like it used to be..

Fixed and tested with flatpak-external-data-checker to be sure it is working.

TingPing: Just double escaping things because of JSON, thanks for testing.

PR #287, opened 2024-01-19
Remove qt4

TingPing:
This hasn't been touched in over 3 years, nobody is an assigned maintainer, and a quick search on Github doesn't show a single package using it.

Since this is so complicated it likely doesn't even work anymore and is time to be removed.


hfiguiere:
This packages uses it

https://github.com/flathub/org.scantailor.ScanTailor/blob/master/org.scantailor.ScanTailor.yaml#L19

hfiguiere:
and it's built on 23.08

TingPing:
Interesting that it still works. However if we have one user maybe it should be moved there. It really is just unmaintained here anyway.

TingPing:
CC @Unrud

Unrud:
I have no objection.

PR #288, opened 2024-01-25
Update to libdecor 0.2.2, rename module

kowalski7cc:


Erick555:
is is still relevant? libdecor is in latest runtimes

bbhtt:
It should be removed when 22.08 goes out of support.

PR #292, opened 2024-02-13
switch libmad to Tenacity fork

Be-ing:
This fork replaces the autotools build system with CMake and incorporates patches that have been floating around distros' packages for years, including the patches that were used here.


bbhtt:
Cleanup: `/include, /lib/cmake, /lib/pkgconfig`

Be-ing:
Thanks for the quick review and merge.

PR #293, opened 2024-02-15
Update libsecret to 0.21.3

ZVNexus:


libsecret/libsecret.json
```diff
@@ -2,6 +2,7 @@
   "name": "libsecret",
   "buildsystem": "meson",
   "config-opts": [
+    "-Dcrypto=gnutls",
```
bbhtt, line 5: This needs gnutls >= 3.8.2, the runtime ships 3.8.0 it silently falls back to libgcrypt.

bbhtt: https://gitlab.com/freedesktop-sdk/freedesktop-sdk/-/merge_requests/17826

A6GibKm: Please propose this change in a separate MR.

ZVNexus: > Please propose this change in a separate PR.

Done

PR #294, opened 2024-02-16
Use GnuTLS as the crypto backend for libsecret

AKoskovich:


ZVNexus:
> This needs gnutls >= 3.8.2, the runtime ships 3.8.0 it silently falls back to libgcrypt.
> 
> Even if it is updatable in runtime, it's not going to reach Flathub before next month. Also 22.08 can never update to 3.8 branch.
> 
> https://gitlab.com/freedesktop-sdk/freedesktop-sdk/-/merge_requests/17826

TingPing:
>  the runtime ships 3.8.0 it silently falls back to libgcrypt.

This means that `-Dcrypto=gnutls` is silently ignored by libsecret? That sounds wrong.

bbhtt:
It's at least not exiting when it can't find the required version of gnutls 3.8.2

Erick555:
> This means that -Dcrypto=gnutls is silently ignored by libsecret? That sounds wrong.

Actually this may be advantage here otherwise it will stop working with fdsdk 22.08 which as noted won't have relevant gnutls.

bbhtt:
As @Erick555 noted in #freedesktop-sdk, updating gnutls to 3.8.2 in 23.08 branch of freedesktop-sdk will break forward ABI guarantees.

So this needs to wait until 24.08 comes out this October or November. The update to `master` was pushed yesterday https://gitlab.com/freedesktop-sdk/freedesktop-sdk/-/merge_requests/17825

TingPing:
> It's at least not exiting when it can't find the required version of gnutls 3.8.2

It just disables *all* crypto:

https://gitlab.gnome.org/GNOME/libsecret/-/blob/master/meson.build?ref_type=heads#L51-62

bbhtt:
Then it can't land in any case, until every supported runtime has the dependency.

TingPing:
https://gitlab.gnome.org/GNOME/libsecret/-/merge_requests/137

A6GibKm:
Just to be clear, if you disable crypto it only means apps will use the dbus backend rather than the secrets portal, it does not disable encryption at the library level.

xanscale:
i think the solution is to update runtime not force a behavior

bbhtt:
If there is no migration path, then this needs to be a separate manifest `libsecret-gnutls.json` so that existing apps don't suddenly break.

Those wanting to use that one can explicitly add it after assessing it for their app.

bbhtt:
Libsecret was added to freedesktop-sdk for 25.08 (default build using gcrypt backend as it will replace this shared module). The current shared modules has no maintainer and will most likely be removed once 25.08 is out.

Chromium and Electron does not work with the default build, they need the DBus backends. I switched chromium to that already https://github.com/flathub/org.chromium.Chromium.BaseApp/commit/3937abe5460d8f47d9a9035b258b8d0048ef42fb

Electron baseapp will be switched to that in the 25.08 branch.

PR #295, opened 2024-02-17
Disable gcrypt and transport encryption for libsecret

ZVNexus:


A6GibKm:
This is not a good idea, it disables the portal backend of libsecret. Not only it will require a sandbox hole (for org.freedesktop.secrets) on the apps relying on this, it will also look (from the point of view of the app) like all the previosly stored secrets were deleted.

This bug https://gitlab.gnome.org/GNOME/libsecret/-/issues/58 might explain why the portal backend does not work for some apps though.

ZVNexus:
> This bug https://gitlab.gnome.org/GNOME/libsecret/-/issues/58 might explain why the portal backend does not work for some apps though.

Is there something that can be done for Slack in the interim? If this requires a fix in libsecret it could be awhile before Slack works again without reverting the addition of libsecret in the Electron baseapp.

A6GibKm:
You can build the library inside the the manifest of your app, not ideal but not terrible either. I am working on fixing the issue at the moment.

ZVNexus:
> You can build the library inside the the manifest of your app, not ideal but not terrible either. I am working on fixing the issue at the moment.

FYI I tested the patch for https://gitlab.gnome.org/GNOME/libsecret/-/merge_requests/138 on Slack, and still running into the issue where user gets signed out every launch so doesn't seem to be the issue for Slack's case.

PR #296, opened 2024-02-23
Update libsecret to 0.21.4

A6GibKm:


A6GibKm:
Fixes https://gitlab.gnome.org/GNOME/libsecret/-/issues/58. But I am not sure what the impact is for apps which might not be using libsecret api directly.

PR #297, opened 2024-02-26
libusb: Update to 1.0.27

A6GibKm:
Tested:

- [x] compiles
- [x] apps run as intended


RobLoach:
Quick note that we're using this in org.libretro.RetroArch, and it seems to be working fine.

A6GibKm:
Thanks for the feedback.

PR #303, opened 2024-05-02
Update SDL2 to 2.30.3

ColinKinloch:


bbhtt:
I'm merging this but this will be dropped in ~2 months when freedesktop sdk 22.08 is EOL. 23.08+ has libdecor, sdl2 and sdl2 with libdecor support.

PR #304, opened 2024-05-10
Extract zmusic and game-music-emu modules to be reused in Raze

smirnov:
Due to addition org.zdoom.Raze engine to flathub it would make sense to extract game-music-emu and zmusic modules into their own files for easier reuse across both project (modules are already defined independently anyway).

### DIFF
--- a/gzdoom/game-music-emu.json
+++ b/gzdoom/game-music-emu.json
@@ -1,0 +1,20 @@
+{
+    "name": "game-music-emu",
+    "buildsystem": "cmake-ninja",
+    "config-opts": [
+        "-DCMAKE_BUILD_TYPE=RelWithDebInfo"
+    ],
+    "sources": [
+        {
+            "type": "archive",
+            "url": "https://github.com/libgme/game-music-emu/archive/refs/tags/0.6.3.tar.gz",
+            "sha256": "4c5a7614acaea44e5cb1423817d2889deb82674ddbc4e3e1291614304b86fca0",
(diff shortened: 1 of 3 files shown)


Eonfge:
Good idea. The only real I would change, is moving the 'gme' and 'zmusic' package to the 'linux-audio' folder.

hfiguiere:
> Good idea. The only real I would change, is moving the 'gme' and 'zmusic' package to the 'linux-audio' folder.

NO.

Eonfge:
@hfiguiere care RI comment? 'gme' and 'zmusic' are also used by other applications not gzdoom related.

hfiguiere:
I maintain that section, so no. Feel free to keep it where it is.

smirnov:
I would also tend to agree to keep it in zdoom, as Raze is effectively a GZDoom fork for Build games, so it's more of a zdoom library, than a linux-sound library, but are there any objections to extracting those modules and merging this PR as is?

Eonfge:
@hfiguiere Fair reason. Was not aware that it affected the permissions on the libraries that you maintain.

@smirnov Good to go then. Let's bundle our forces with zmusic and gme

Eonfge (approved):

Eonfge:
Are we waiting on me to merge it?

smirnov:
Well, I have no ability to merge it, so if you could, that would be great.

Eonfge:
:saluting_face:

PR #308, opened 2024-08-09
Build GZDoom without Debug Info

Eonfge:
https://github.com/ZDoom/gzdoom/issues/2656


Erick555:
Note that flatpak-builder will still apply -g -Wp,-D_GLIBCXX_ASSERTIONS from C/CXXFLAGS environment.

cncz42:
I was about to comment the same discovery lol, doing a fresh build and relwithdebinfo alone doesn't trigger the crash. Looks like it was similar to the makepkg issue I had earlier, probably no need to remove the debug info although idk how it affects performance

Erick555:
`Relwithdebinfo` shouldn't affect performance at all - it just add extra information to the binaries which is then stripped to separate extension. There is separate `Debug` mode which is the one that may affect perfromance.

https://github.com/ZDoom/gzdoom/blob/master/CMakeLists.txt#L153

cncz42:
@Erick555 how would you go about overriding the default args via the manifest?

Erick555:
one option is to set [CXXFLAGS](https://github.com/flathub/org.freedesktop.Sdk.Extension.llvm18/blob/branch/23.08/org.freedesktop.Sdk.Extension.llvm18.json#L51)  env variable.

PR #309, opened 2024-08-12
Remove assertions on GZDoom

cncz42:
@Eonfge 
Fully fixes https://github.com/ZDoom/gzdoom/issues/2656.

### DIFF
--- a/gzdoom/gzdoom.json
+++ b/gzdoom/gzdoom.json
@@ -13,7 +13,7 @@
         "/lib/*.la",
         "/lib/pkgconfig",
         "/include"
-    ]
+    ],
     "sources": [
         {
             "type": "git",
--- a/gzdoom/gzdoom.json
+++ b/gzdoom/gzdoom.json
@@ -4,15 +4,16 @@
     "config-opts": [
         "-DCMAKE_BUILD_TYPE=Release"
     ],
+    "build-options" : {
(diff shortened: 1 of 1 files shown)


Eonfge:
I validated your changes and they are good. I'll hold of until Upstream confirms that it;
https://github.com/ZDoom/gzdoom/issues/2656#issuecomment-2283161305

gzdoom/gzdoom.json
```diff
@@ -4,6 +4,9 @@
     "config-opts": [
         "-DCMAKE_BUILD_TYPE=Release"
     ],
+    "build-options" : {
+        "cxxflags": "-Wp,-U_GLIBCXX_ASSERTIONS"
```
hfiguiere, line 8: if the code crashes with these assertions the code should be fixed. These assertions exists for a reason 100% of their triggers are bug that are usually obscured.

Eonfge: I'm discussing the issue Upstream. If they choose to distribute GZDoom with these known issues, then I don't think it's my job to decide otherwise. I'll try to keep this Flathub as close as possible to the official GZDoom binary.

Eonfge:
Forgot to close this issue. Ultimately, I took the patched code from upstream to fix this issue. With the next release of GZDoom, the patch will become unnecessary.

PR #319, opened 2024-09-29
Add libayatana-appindicator

ZVNexus:
Currently there are over a dozen apps using this package now.

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -21,3 +21,4 @@
 /luajit/           @ranisalt
 /lzo/              @ZVNexus
 /squashfs-tools/   @ZVNexus
+/libayatana-appindicator/  @ZVNexus
--- a/libayatana-appindicator/libayatana-appindicator-gtk3.json
+++ b/libayatana-appindicator/libayatana-appindicator-gtk3.json
@@ -1,0 +1,118 @@
+{
+    "name": "libayatana-appindicator",
+    "buildsystem": "cmake-ninja",
+    "cleanup": [
+        "/include",
+        "/lib/girepository-1.0",
+        "/lib/pkgconfig",
+        "/share/gir-1.0"
+    ],
+    "config-opts": [
+        "-DENABLE_BINDINGS_MONO=NO",
+        "-DENABLE_BINDINGS_VALA=NO",
+        "-DENABLE_GTKDOC=NO"
+    ],
+    "modules": [
+        "../intltool/intltool-0.51.json",
+        {
+            "name": "libdbusmenu",
+            "buildsystem": "autotools",
+            "build-options": {
+                "cflags": "-Wno-error"
+            },
+            "cleanup": [
+                "*.la",
+                "/include",
+                "/lib/pkgconfig",
+                "/libexec",
+                "/share/doc",
+                "/share/gtk-doc"
+            ],
+            "config-opts": [
+                "--with-gtk=3",
+                "--disable-dumper",
+                "--disable-static",
+                "--disable-tests",
+                "--disable-gtk-doc",
+                "--enable-introspection=no",
+                "--disable-vala"
+            ],
+            "sources": [
+                {
+                    "type": "archive",
+                    "url": "https://launchpad.net/libdbusmenu/16.04/16.04.0/+download/libdbusmenu-16.04.0.tar.gz",
+                    "sha256": "b9cc4a2acd74509435892823607d966d424bd9ad5d0b00938f27240a1bfa878a"
+                },
+                {
+                    "type": "patch",
+                    "paths": [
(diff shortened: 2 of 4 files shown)


libayatana-appindicator/libayatana-appindicator-gtk3.json
```diff
@@ -0,0 +1,101 @@
...
+                {
+                    "type": "shell",
+                    "//": "Remove unused Vala/GIR files",
+                    "commands": [
+                        "sed -i 's/Vala REQUIRED/Vala /' src/CMakeLists.txt",
+                        "sed -i '/install.*AyatanaIdo3-0.4.*/d' src/CMakeLists.txt"
+                    ]
+                }
+            ]
+        },
```
bbhtt, line 71: this should be a patch

CODEOWNERS
bbhtt, line 1: can you remove the indentation changes from here ?

libayatana-appindicator/libayatana-appindicator-gtk3.json
```diff
@@ -0,0 +1,101 @@
...
+                    "type": "archive",
+                    "url": "https://launchpad.net/libdbusmenu/16.04/16.04.0/+download/libdbusmenu-16.04.0.tar.gz",
+                    "sha256": "b9cc4a2acd74509435892823607d966d424bd9ad5d0b00938f27240a1bfa878a"
+                }
+            ],
+            "cleanup": [
+                "/include",
+                "/libexec",
+                "/lib/pkgconfig",
+                "/lib/*.la",
```
bbhtt, line 41:
```suggestion
                "*.la",
```

libayatana-appindicator/libayatana-appindicator-gtk3.json
```diff
@@ -0,0 +1,101 @@
...
+    "modules": [
+        "../intltool/intltool-0.51.json",
+        {
+            "name": "libdbusmenu",
+            "buildsystem": "autotools",
+            "build-options": {
+                "cflags": "-Wno-error",
+                "env": {
+                    "HAVE_VALGRIND_FALSE": "#",
+                    "HAVE_VALGRIND_TRUE": ""
```
hfiguiere, line 18: no. this should be a config-opt. But it seems that `--disable-tests` does the trick.

does that even work?

libayatana-appindicator/libayatana-appindicator-gtk3.json
hfiguiere: and the patch shoud be submitted upstream to remove the `REQUIRED`.

libayatana-appindicator/libayatana-appindicator-gtk3.json
```diff
@@ -0,0 +1,101 @@
...
+                        "tag-pattern": "^([\\d.]+)$"
+                    }
+                }
+            ]
+        }
+    ],
+    "sources": [
+        {
+            "type": "git",
+            "url": "https://github.com/AyatanaIndicators/libayatana-appindicator.git",
```
hfiguiere, line 92: I don't understand why there are two module with different tag but same repo?

libayatana-appindicator/libayatana-appindicator-gtk3.json
ZVNexus: This was required because disable-tests means HAVE_VALGRIND never gets set, needs a patch to fix that. But libdbusmenu is dead upstream so not really any chance of getting the patch in.

libayatana-appindicator/libayatana-appindicator-gtk3.json
ZVNexus: Fixed

CODEOWNERS
ZVNexus: Fixed, but is it alright to have the indentation be inconsistent?

libayatana-appindicator/libayatana-appindicator-gtk3.json
ZVNexus: Removing the required isn't enough, since find_package still gets run and it fails. Just added a patch to not bother with AyatanaIdo3-0.4.gir if ENABLE_INTROSPECTION is disabled.

bbhtt:
You have to add cleanups for the following:

```
/include/libayatana-appindicator3-0.1
/lib/pkgconfig/ayatana-appindicator3-0.1.pc
```

ZVNexus:
> You have to add cleanups for the following:
> 
Fixed

ZVNexus:
Changed the ayatana-ido patch to remove the indentation changes since it'll conflict with upstream merges.

bbhtt:
Looks like the typelib files are needed to interact via Python, I will remove the cleanup.

PR #323, opened 2024-09-30
lzo: Cleanup /lib/pkgconfig/lzo2.pc

ZVNexus:
Also don't cleanup /lib/*.so, we're just cleaning up a symlink to liblzo2.so.2.0.0 and as an independent module it should be shared.

### DIFF
--- a/lzo/lzo.json
+++ b/lzo/lzo.json
@@ -2,7 +2,7 @@
     "name": "lzo",
     "cleanup": [
         "/include",
-        "/lib/*.so",
+        "/lib/pkgconfig",
         "/share/doc",
         "*.la"
     ],


Erick555:
What do you mean by _should be shared._ It's cleaned only after everything is build.

ZVNexus:
> Collaborator

It's a shared library for squashfs-tools, `"/lib/*.so",` only removes the liblzo2.so symlink, not liblzo2.so.2 & liblzo2.so.2.0.0 which is required for unsquashfs so just remove cleanup `"/lib/*.so",` all together since liblzo2.so is just a symlink.

(unsquashfs has a runtime dependency on liblzo2.so.2, why they're not cleaned)

Erick555:
the entire point is to remove that symlink because it's build time thing. runtime does this for all symlinks

PR #332, opened 2024-10-13
mac: Set CMake install dirs

enzo1982:
Without this, libMAC is installed in lib64 when building for x86-64 with Freedesktop SDK 24.08.

### DIFF
--- a/mac/mac.json
+++ b/mac/mac.json
@@ -1,6 +1,11 @@
 {
   "name": "libmac",
   "buildsystem": "cmake",
+  "config-opts": [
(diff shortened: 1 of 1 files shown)


bbhtt:
That's not the issue, this is set via flatpak-builder. For 24.08 you have to use flatpak-builder 1.4.4

bbhtt:
it's mentioned in the release notes https://discourse.flathub.org/t/freedesktop-sdk-24-08-0-released/7630/4

enzo1982:
Thanks @bbhtt! Confirmed that it works with flatpak-builder 1.4.4.

Closing this.

Issue #345, opened 2024-11-09
Why squashfs-tools no zstd support ?


Boria138:

hfiguiere:
issue description is empty

Boria138:
As I asked in the title, why squashfs-tools is built without ZSTD_SUPPORT, mksquashfs -comp zstd doesn't work without it, also GZIP_SUPPORT=1 LZ4_SUPPORT=1 LZMA_XZ_SUPPORT=1 XATTR_SUPPORT=1 is missing, I would add gzip and zstd as the most popular compression methods to the already existing lzo and xz ones

bbhtt:
You can open a PR it is missing because no one added it.

PR #346, opened 2024-11-09
Enable ZSTD_SUPPORT and GZIP_SUPPORT on squashfs-tools

Boria138:
see https://github.com/flathub/shared-modules/issues/345

### DIFF
--- a/squashfs-tools/squashfs-tools.json
+++ b/squashfs-tools/squashfs-tools.json
@@ -2,7 +2,7 @@
     "name": "squashfs-tools",
     "buildsystem": "simple",
     "build-commands": [
-        "XZ_SUPPORT=1 LZO_SUPPORT=1 make -C squashfs-tools -j ${FLATPAK_BUILDER_N_JOBS} install INSTALL_DIR=${FLATPAK_DEST}/bin INSTALL_MANPAGES_DIR=${FLATPAK_DEST}/man/man1"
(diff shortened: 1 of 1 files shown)


squashfs-tools/squashfs-tools.json
```diff
@@ -1,8 +1,26 @@
+{
+    "name": "zstd",
+    "buildsystem": "meson",
+    "builddir": "true"
+    "subdir": "build/meson",
+    "sources": [
+        {
+            "type": "archive",
+            "url": "https://github.com/facebook/zstd/releases/download/v1.5.6/zstd-1.5.6.tar.gz",
```
hfiguiere, line 9: this is in Freedesktop SDK 24.08. And 1.5.5 is in 23.08

So you don't need it.

Boria138: Drop zstd build, thanks

squashfs-tools/squashfs-tools.json
```diff
@@ -2,7 +2,7 @@
     "name": "squashfs-tools",
     "buildsystem": "simple",
     "build-commands": [
-        "XZ_SUPPORT=1 LZO_SUPPORT=1 make -C squashfs-tools -j ${FLATPAK_BUILDER_N_JOBS} install INSTALL_DIR=${FLATPAK_DEST}/bin INSTALL_MANPAGES_DIR=${FLATPAK_DEST}/man/man1"
+        "XZ_SUPPORT=1 LZO_SUPPORT=1 ZSTD_SUPPORT=1 GZIP_SUPPORT=1 make -C squashfs-tools -j ${FLATPAK_BUILDER_N_JOBS} install INSTALL_DIR=${FLATPAK_DEST}/bin INSTALL_MANPAGES_DIR=${FLATPAK_DEST}/man/man1"
```
bbhtt, line 5: Not related to your PR but this should not use `simple`,  `make-args` and `make-install-args` exist with `no-configure`.

PR #363, opened 2025-01-11
Fix gzdoom crashing certain mods

nvllsvm:
For example, the mod "Doom 64: Unseen Evil (v1.03)" crashes when attempting to start a level without this fix.

Resolves https://github.com/ZDoom/gzdoom/issues/2834

### DIFF
--- a/gzdoom/zmusic.json
+++ b/gzdoom/zmusic.json
@@ -1,6 +1,9 @@
 {
     "name": "zmusic",
     "buildsystem": "cmake-ninja",
+    "build-options": {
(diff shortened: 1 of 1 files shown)


nvllsvm:
I found out about this after opening this PR: https://github.com/Chainmanner/GZDoom-Arbitrary-Code-Execution-via-ZScript-PoC/issues/2

This _probably_ should be merged.

hfiguiere:
just to be clear, disabling this assertion is NEVER a solution.

Eonfge:
As GZDoom package maintainer, I would prefer Upstream to fix their type castings. GZDoom had some security issues recently with weak type checking, so I'm against such a short-term fix.

PR #379, opened 2025-03-19
libusb: Update to v1.0.28

A6GibKm:


A6GibKm:
Anitya still thinks 1.0.28.rc is the latest one so we might want to wait a few days.

A6GibKm:
It should be safe to merge now.

PR #384, opened 2025-04-18
Update libusb

tsdgeos:


tsdgeos:
ok, duplicated work https://github.com/flathub/shared-modules/pull/379

PR #401, opened 2025-08-25
Add SDL2-image and SDL2-ttf

tothxa:
These will be removed from the runtime in 25.08, but are likely required by many SDL2 apps. Feel free to postpone until the new runtime is released.

Tested at: https://github.com/flathub/org.widelands.Widelands/pull/80

I didn't add myself to `CODEOWNERS` because as I see it the `SDL1.2` modules don't have a maintainer either, and I don't think I'd be the right person.

Actually, I just took the old modules from the `SDL` directory and updated the versions. They just worked as far as I can tell.

### DIFF
--- a/SDL2/SDL2_image.json
+++ b/SDL2/SDL2_image.json
@@ -1,6 +1,8 @@
 {
     "name": "SDL2_image",
-    "config-opts": ["--disable-static"],
+    "config-opts": [
+        "--disable-static"
+    ],
     "rm-configure": true,
     "cleanup": [
         "/include",
@@ -10,9 +12,10 @@
     ],
     "sources": [
         {
-            "type": "archive",
-            "url": "https://www.libsdl.org/projects/SDL_image/release/SDL2_image-2.8.8.tar.gz",
-            "sha256": "2213b56fdaff2220d0e38c8e420cbe1a83c87374190cba8c70af2156097ce30a",
+            "type": "git",
+            "url": "https://github.com/libsdl-org/SDL_image.git",
+            "tag": "release-2.8.8",
+            "commit": "c1bf2245b0ba63a25afe2f8574d305feca25af77",
             "x-checker-data": {
                 "type": "git",
                 "tag-pattern": "^release-(2\\.[\\d.]+)$"
--- a/SDL2/SDL2_image.json
+++ b/SDL2/SDL2_image.json
@@ -3,30 +3,27 @@
     "config-opts": [
         "--disable-static"
     ],
-    "rm-configure": true,
     "cleanup": [
         "/include",
+        "/lib/cmake",
         "/lib/pkgconfig",
         "*.la",
         "*.a"
     ],
     "sources": [
         {
-            "type": "git",
-            "url": "https://github.com/libsdl-org/SDL_image.git",
-            "tag": "release-2.8.8",
-            "commit": "c1bf2245b0ba63a25afe2f8574d305feca25af77",
+            "type": "archive",
+            "url": "https://www.libsdl.org/projects/SDL_image/release/SDL2_image-2.8.8.tar.gz",
+            "sha256": "2213b56fdaff2220d0e38c8e420cbe1a83c87374190cba8c70af2156097ce30a",
             "x-checker-data": {
-                "type": "git",
-                "tag-pattern": "^release-(2\\.[\\d.]+)$"
+                "type": "anitya",
+                "project-id": 382729,
+                "stable-only": true,
+                "versions": {
+                    "<": "3.0.0"
+                },
+                "url-template": "https://www.libsdl.org/projects/SDL_image/release/SDL2_image-$version.tar.gz"
             }
-        },
-        {
-            "type": "script",
-            "dest-filename": "autogen.sh",
-            "commands": [
-                "AUTOMAKE=\"automake --foreign\" autoreconf -vfi"
-            ]
         }
     ]
 }
--- a/SDL2/SDL2_image.json
+++ b/SDL2/SDL2_image.json
@@ -1,0 +1,25 @@
+{
+    "name": "SDL2_image",
+    "config-opts": ["--disable-static"],
+    "rm-configure": true,
+    "cleanup": [
+        "/include",
+        "/lib/pkgconfig",
+        "*.la",
+        "*.a"
+    ],
+    "sources": [
+        {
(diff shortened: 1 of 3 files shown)


bbhtt:
Can you rebase to master?

tothxa:
Oops, how stupid of me...

However anitya for SDL2_ttf seems to be wrong: https://release-monitoring.org/project/4784/
(SDL2_image does have its separate [entry](https://release-monitoring.org/project/382729/) though)

Would it be OK to use git for the sources instead? Like:
```
    "sources": [
        {
            "type": "git",
            "url": "https://github.com/libsdl-org/SDL_image.git",
            "tag": "release-2.8.8",
            "commit": "c1bf2245b0ba63a25afe2f8574d305feca25af77",
            "x-checker-data": {
                "type": "git",
                "tag-pattern": "^release-(2\\.[\\d.]+)$"
            }
        },
```

bbhtt:
yea that's fine

v1993:
Would it make sense to switch to CMake + Ninja instead of autoconf since these libraries fully support it as a build system? That should make build recipe a bit cleaner.

bbhtt:
Good idea, @tothxa what do you think?

user1-github:
I already [ship SDL2_image](https://github.com/flathub/net.redeclipse.RedEclipse-Legacy/blob/8cb204844d47e6418c98985ce9b55a488bdfffa6/net.redeclipse.RedEclipse-Legacy.yml#L22) with a game I maintain and in my case it builds and works just fine without specifying buildsystem or any other options.
I also suggest adding /lib/cmake to cleanup section.

user1-github:
Then you can leave --disable-static to config-opts (but you have .a in cleanup section anyway).

I just thought they wanted to switch to cmake-ninja in order to avoid needing these lines:

>             "type": "script",
            "dest-filename": "autogen.sh",
            "commands": [
                "AUTOMAKE=\"automake --foreign\" autoreconf -vfi"

so I wanted to show that simply removing them is enough.

Erick555:
autotools is legacy buildsystem, if upstream support cmake/meson it's better to use that.

tothxa:
OK, switched to `cmake`, thanks for the tip. As I said, I just took the SDL1.2 files, threw them at the wall, and they stuck. :)

BTW @user1-github @Erick555 Doesn't either of you want to take this over? You seem far more qualified than me.

tothxa:
@bbhtt Looks like git needs to download too much stuff from too many places, so I fought `anitya` after all...

user1-github:
@tothxa Just add `"buildsystem": "cmake-ninja"` under name.

Btw, you don't need to specify 
`"versions": {
         "<": "3.0.0"`
for SDL2_image, as id 382729 is [dedicated](https://release-monitoring.org/project/382729/) specifically for SDL2_image.

Unfortunately idk what maintaining shared modules involve. For me personally it's also much more convenient to just ship a dependency in my app manifest, unless it's a very old eol dependency that doesn't get updated anymore.

tothxa:
> Btw, you don't need to specify `"versions": { "<": "3.0.0"` for SDL2_image, as id 382729 is [dedicated](https://release-monitoring.org/project/382729/) specifically for SDL2_image.

I know, it was just meant for extra safety. But OK, I removed it, as well as the not built static libs from cleanup.

> If you're using anitya, you also need to change the source url and url template to Github, as [www.libsdl.org](http://www.libsdl.org) urls won't work with that. See the link to my game manifest I provided in my first comment as an example.

What do you mean they won't work? I've tested it (changed the versions to some old ones and changed the sha256 randomly, then ran `external-data-checker --update --edit-only` on them, and it updated everything back just fine. OK, maybe anitya checks the github artifacts, and libsdl.org may be updated later after a release, but I don't think that should be a big issue, and libsdl.org sounds more official...

SDL2/SDL2_image.json
```diff
@@ -0,0 +1,23 @@
+{
+    "name": "SDL2_image",
+    "cleanup": [
+        "/include",
+        "/lib64/cmake",
+        "/lib64/pkgconfig"
```
bbhtt, line 6:
```suggestion
        "/lib/cmake",
        "/lib/pkgconfig"
```

tothxa: I tested it locally, and the cmake build puts these in `lib64`.

tothxa:
> 1. Would you mind also adding SDL2_mixer? If not, I guess I could PR it separately once this is merged. I [got it working](https://github.com/flathub/io.github.lincity_ng.lincity-ng/blob/f491d7d63f11c81f39baae63a602d6a655a04ce9/SDL2/SDL2_mixer.json) without much trouble.
>
>     2. There's also SDL2_net in this family of libraries, but I personally don't have a project that uses it either, so perhaps it's best left for someone else to submit.

As far as I can tell, SDL3_mixer and SDL3_net are not yet released upstream, so their SDL2 versions are in the new runtime. We use SDL2_mixer too, and my test build succeeded without adding it as a module.

>     3. It might be worth checking https://gitlab.com/freedesktop-sdk/freedesktop-sdk/-/tree/release/24.08/elements/components to see what flags were previously passed when building these libraries to ensure minimal breakage when migrating. I found it helpful for getting SDL2_mixer working and it looks like some format-related options were set for SDL2_image.

I've thought of that. E.g. it's strange that using the systemwide format libraries is disabled for SDL_image... it would be nice to know the reason, as I assume it must have been important at the time...
But this is already turning into much more effort than what I was prepared for, so I don't want to track it down, and I'm reluctant to make it less efficient without knowing why.

>     4. It's probably still a good idea to switch to Github for downloads even if the current path works with Anitya, since libsdl.org downloads are seemingly deprecated, with everything pointing at the corresponding Github releases pages instead.

I see. Indeed, the releases link on libsdl.org points to github. Will do it.

v1993:
> You may see that all image formats are set to `ON` by [default](https://github.com/libsdl-org/SDL_image/blob/release-2.8.8/CMakeLists.txt#L84-L100)

I see. Looks like this was changed relatively recently in https://github.com/libsdl-org/SDL_image/commit/542c405c29626b1eac5fd5431ca0e90f10eec4ff. In this case, perhaps enable the related `SDL2IMAGE_STRICT` so that missing formats aren't silently skipped during the build?

tothxa:
Thanks for the investigation and explanation.

PR #415, opened 2025-09-07
Use CMake 3.5

Eonfge:


Eonfge:
Tested it with Asunder and I didn't find any anomalies

Erick555:
The commit and PR title should say which project it touches.

PR #419, opened 2025-09-26
physfs: Update to 3.2.0 and fix building with FreeDesktop SDK 25.08

ProjectSynchro:
Passing in the config-opt: `"-DCMAKE_POLICY_VERSION_MINIMUM=3.5"` solves the build issue observed when using CMake 4.

I also went ahead and updated the version of physfs to 3.2.0. (up from 3.0.2) It looks like they have moved to Github from their old release location. 

I added an x-checker-data check for future updates which should help with the version falling behind again.

### DIFF
--- a/physfs/physfs.json
+++ b/physfs/physfs.json
@@ -1,7 +1,7 @@
 {
     "name": "PhysicsFS",
     "buildsystem": "cmake-ninja",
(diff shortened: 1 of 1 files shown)


Mailaender (approved):

Mailaender:
Thanks!

PR #435, opened 2025-12-12
clutter: Fix cogl build

hfiguiere:
Force gnu18 standard as it fails in C23 the new default


hfiguiere:
blocks https://github.com/flathub/org.gnome.TwentyFortyEight/pull/15

PR #452, opened 2026-03-11
Add SDL2_mixer

user1-github:
This should be merged towards SDK 26.08 release.

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -20,4 +20,4 @@ update-checker.sh  @flathub/reviewers
 /lzo/              @AKoskovich
 /squashfs-tools/   @AKoskovich
 /libayatana-appindicator/  @AKoskovich
-/SDL2/             @tothxa
+/SDL2/             @tothxa @user1-github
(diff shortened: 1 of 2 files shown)


user1-github:
@tothxa Since you're already a codeowner of SDL2 namespace, I think you should be able to merge this.

tothxa:
OK, just let me run a test first, please.

Thanks for the heads-up about the release of 26.08.

tothxa (approved):

tothxa:
Sorry, it looks like I can't merge for some reason after all.

user1-github:
Then I guess only Flathub admins can do it.
But the one who I remember was active in this repo has recently left.

@barthalion Maybe you can merge this? Idk who else is responsible for shared modules repo.

tothxa:
I think the problem is that I'm only a code owner for the `SDL2` directory, but changing the `CODEOWNERS` file needs higher privileges, even if you only try to add yourself to this directory.

PR #462, opened 2026-05-10
SDL2_image: use libjpeg and libpng instead of stb_image

s09bQ5:
stb_image is a public domain implementation of several image decoders in a single C header file. SDL2_image includes a copy of it and uses it by default for JPEG and PNG. A comment in stb_image says

  "Primarily of interest to game developers and other people who can
   avoid problematic images and only need the trivial interface"

Since the flatpak runtimes generally do include libjpeg and libpng, there is no reason to use a possibly limited implementation. It probably also receives less security testing than the de-facto standard libraries.

### DIFF
--- a/SDL2/SDL2_image.json
+++ b/SDL2/SDL2_image.json
@@ -1,6 +1,7 @@
 {
     "name": "SDL2_image",
     "config-opts": [
+        "-DSDL2IMAGE_BACKEND_STB=OFF",
         "-DSDL2IMAGE_DEPS_SHARED=OFF",
         "-DSDL2IMAGE_STRICT=ON"
     ],


tothxa:
I don't really know what to think about this. Maybe I'm too old, but "limited implementation with trivial interface" sounds like the essence of security to me. (but I can't judge the actual code myself)

As for its limits: We use somewhat esoteric PNGs to save space (indexed colours with separate full alpha) that most image editors can't edit and save properly, but they work just fine with the current SDL2-image builds. 

And SDL itself is "Primarily of interest to game developers". :)

tothxa:
@s09bQ5 I've just read your linked comment. Have you tried building UltraStarDeluxe with including the current manifest? If so, are you aware of any actual image files that didn't work with that build?

s09bQ5:
Our game requires users to add content, so we don't have any control over the images.

LaxentaInc:
HAII guys ikr yall are linux users but check out https://colorwall.xyz its a desktop engine in rust for windows tho, i am just asking for opinions, no hard feelings @s09bQ5 @tothxa @ebassi @cosimoc @Grumbel

PR #475, opened 2026-07-07
linux-audio: Update fluidsynth 2.5.6

hfiguiere:
Address several CVE


hfiguiere:
tested with a package.

PR #491, opened 2026-09-04
New module: appstream compose

hfiguiere:
This is necessary for at least half a dozen packages since appstreamcli was removed from the 26.08 SDK. (ref https://gitlab.com/freedesktop-sdk/freedesktop-sdk/-/work_items/1906 )

appstreamcli compose is needed to compose the appstream file for bundled extension that are not in the same app-id prefix, in that case audio plugins.

A more ideal solution would be to implement this case in flatpak-builder since it know about the bundle extension it could do the compose. But this will take longer.

I added myself to the CODEOWNER.


appstream/appstream-compose.json
```diff
@@ -0,0 +1,129 @@
...
+                {
+                    "type": "git",
+                    "url": "https://github.com/hughsie/libxmlb.git",
+                    "commit": "d942cd68a7d78e35b312b9299e69f68f9cd2e503",
+                    "tag": "0.3.29"
+                }
+            ]
+        },
+        {
+            "name": "blake3",
```
Erick555: you may just pass `-Dblake3-support=false` in appstream build

hfiguiere: I thought about it but:

1. it is for compose (not sure exactly what)
2. this should be the same behaviour as upstream.

Maybe i'll just pretend we are an old distro.

hfiguiere: done

CodedOre (approved):
LGTM

